PR #1, opened 2022-09-30
Added new FeralFileArtworkV3 contract

tienngovan:


lemonlatte (changes requested):
Could you update the indent use 2 space?

For `migrations/3_feralfile_exhibition_v2.js`, it seems only indent changes?

contracts/FeralfileArtworkV3.sol
```diff
@@ -0,0 +1,433 @@
...
+        receiver = royaltyPayoutAddress;
+
+        royaltyAmount =
+            (salePrice * secondarySaleRoyaltyBPS) /
+            MAX_ROYALITY_BPS;
+    }
+
+    /// @notice isValidRequest validates a message by ecrecover to ensure
+    //          it is signed by either the contract owner or a trustee
+    //          who is set by the contract owner.
```
lemonlatte: The signature comes from the token owner, the token owner might not be a trustee.
The scenario here is the trustee execute the transfer base on the consent of the token owner.

contracts/FeralfileArtworkV3.sol
```diff
@@ -0,0 +1,433 @@
...
+        uint8 v_,
+        uint256 tokenID_
+    ) internal view returns (bool authorized) {
+        address signer = ECDSA.recover(
+            ECDSA.toEthSignedMessageHash(message_),
+            v_,
+            r_,
+            s_
+        );
+        authorized =
```
lemonlatte: It supposes to validate the signer is `transferParam_.from`.

lemonlatte (changes requested):
The functions are modularized very well here.

contracts/FeralfileArtworkV3.sol
```diff
@@ -0,0 +1,433 @@
...
+        require(
+            artworks[artworkID_].editionSize > 0,
+            "FeralfileExhibitionV3: artwork is not found"
+        );
+        /// @notice The range of editionNumber should be between 0 (AP) ~ artwork.editionSize
+        require(
+            editionNumber_ <= artworks[artworkID_].editionSize,
+            "FeralfileExhibitionV3: edition number exceed the edition size of the artwork"
+        );
+        require(owner_ != address(0), "invalid owner address");
```
lemonlatte, line 400:
```suggestion
        require(owner_ != address(0), "invalid owner address");
        require(artist_ != address(0), "invalid artist address");
```
If we do not accept artist in zero address.

lemonlatte (changes requested):

contracts/FeralfileArtworkV3.sol
tienngovan: Fixed, changed verify signer is transferParam_.from

contracts/FeralfileArtworkV3.sol
tienngovan: Added

lemonlatte (approved):
LGTM

PR #2, opened 2022-10-07
Changed createArtworks to array

tienngovan:


contracts/FeralfileArtworkV3.sol
```diff
@@ -136,10 +139,10 @@ contract FeralfileExhibitionV3 is ERC721Enumerable, Authorizable, IERC2981 {
...
 
         Artwork memory artwork = Artwork(
-            title = title,
-            artistName = artistName,
-            fingerprint = fingerprint,
-            editionSize = editionSize
+            artwork_.title,
+            artwork_.artistName,
+            artwork_.fingerprint,
+            artwork_.editionSize
```
lemonlatte: Suggest to use keyword argument to prevent the variable position mismatching in the future.

lemonlatte (changes requested):
Is it necessary to create a new `migrations/201_feralfile_exhibition_v3.js`? Can we just use 200?

contracts/FeralfileArtworkV3.sol
tienngovan: Fixed

PR #3, opened 2022-10-11
Updated multiple trustee and removed max edition per artwork

tienngovan:


contracts/Authorizable.sol
```diff
@@ -4,18 +4,23 @@ pragma solidity >=0.4.22 <0.9.0;
 import "@openzeppelin/contracts/access/Ownable.sol";
 
 contract Authorizable is Ownable {
-    address public trustee;
+    address[] public trustees;
```
lemonlatte: Need to be able to set max length of trustee list.

Could be a private function that would call from FFV3 contract consturctor to set an initial number (can not update after deployed).

contracts/Authorizable.sol
```diff
@@ -4,18 +4,23 @@ pragma solidity >=0.4.22 <0.9.0;
...
+                break;
+            }
+        }
+        require(isAuthorized || msg.sender == owner());
         _;
     }
 
-    function setTrustee(address _newTrustee) public onlyOwner {
-        trustee = _newTrustee;
+    function setTrustees(address[] memory _newTrustees) public onlyOwner {
```
lemonlatte: use add / remove seems more flexible.

contracts/Authorizable.sol
```diff
@@ -4,18 +4,23 @@ pragma solidity >=0.4.22 <0.9.0;
...
 
-    constructor() {
-        trustee = address(0x0);
-    }
+    constructor() {}
 
     modifier onlyAuthorized() {
-        require(msg.sender == trustee || msg.sender == owner());
+        bool isAuthorized = false;
+        for (uint256 i = 0; i < trustees.length; i++) {
```
lemonlatte: I prefer using map for `trustees` so that we do not need to validation by loop. Otherwise, the cost of execution would be varied by the index of a trustee.

lemonlatte (changes requested):

contracts/Authorizable.sol
tienngovan: Changed array to map

contracts/Authorizable.sol
tienngovan: Fixed

contracts/Authorizable.sol
tienngovan: Changed array to map

contracts/Authorizable.sol
```diff
@@ -4,18 +4,24 @@ pragma solidity >=0.4.22 <0.9.0;
 import "@openzeppelin/contracts/access/Ownable.sol";
 
 contract Authorizable is Ownable {
-    address public trustee;
+    mapping(address => bool) private trustees;
```
lemonlatte: Trustee can be public. Otherwise, you need to find trustee from the historical transaction. It is relative inconvenient. Is there any security concern that you want to hide it?

contracts/Authorizable.sol
lemonlatte: :+1:

lemonlatte (approved):
LGTM

PR #7, opened 2022-11-07
 Changed timestamp to expire time in authorized transfer

tienngovan:


test/feralfile_exhibition_v3.js
```diff
@@ -297,7 +297,7 @@ contract('FeralfileExhibitionV3', async (accounts) => {
 
     let editionID = BigInt(this.artworkID) + BigInt(editionNumber);
 
-    let timestamp = (new Date().getTime() / 1000 - 400).toFixed(0);
+    let timestamp = (new Date().getTime() / 1000 + 300).toFixed(0);
```
lemonlatte: expiry

test/feralfile_exhibition_v3.js
```diff
@@ -336,7 +336,7 @@ contract('FeralfileExhibitionV3', async (accounts) => {
 
     let editionID = BigInt(this.artworkID) + BigInt(editionNumber);
 
-    let timestamp = (new Date().getTime() / 1000).toFixed(0);
+    let timestamp = (new Date().getTime() / 1000 + 300).toFixed(0);
```
lemonlatte: expiry

contracts/FeralfileArtworkV3.sol
```diff
@@ -342,16 +342,16 @@ contract FeralfileExhibitionV3 is ERC721Enumerable, Authorizable, IERC2981 {
         );
 
         require(
-            transferParam_.timestamp + 5 minutes >= block.timestamp,
-            "FeralfileExhibitionV3: timestamp is over recv window"
+            block.timestamp < transferParam_.expireTime,
```
lemonlatte: `<=` would be better.

lemonlatte (changes requested):

lemonlatte (changes requested):
I think we can move timestamp out of the the `transferParam_` and share between all transactions. This would reduce a lots of gas fee.

```solidity
function authorizedTransfer(expiryTime memory uint256, TransferArtworkParam[] memory transferParams_)
```

The timestamp still need to be part of the signature for each signing.

PR #15, opened 2023-03-09
Implemented MoMA Burn contract

tienngovan:


migrations/205_mint_and_burn.js
```diff
@@ -0,0 +1,9 @@
+var MoMABurn = artifacts.require('MoMABurn');
+
+const argv = require('minimist')(process.argv.slice(2), {
+  string: ['exhibition_curator'],
+});
```
lpopo0856: remove not used

contracts/MoMABurn.sol
```diff
@@ -0,0 +1,136 @@
...
+            );
+
+            if (!isApprovalForAll) {
+                require(
+                    burnContract.getApproved(_burnedEditions[i]) ==
+                        address(this),
+                    "Token need to approve for this contract"
+                );
+            }
+
```
lpopo0856: Logic missing:

Missing check if the token we're going to burn here is actually belongs to the burning artwork. If someone accidentally submit another token which is not the 5k edition MoMA artwork but in same contract to this function we will burn it.

lpopo0856 (changes requested):

contracts/MoMABurn.sol
```diff
@@ -0,0 +1,136 @@
...
+    /// @notice setBurnEnabled use for update burn enable status
+    /// @param _isBurnEnabled - the condition is true or false
+    function setBurnEnabled(bool _isBurnEnabled) public onlyOwner {
+        isBurnEnabled = _isBurnEnabled;
+    }
+
+    /// @notice burnAndMint use for burn old editions and mint new edition
+    /// @param _burnedEditions - array of edition IDs to burn
+    function burnAndMint(uint256[] memory _burnedEditions) public {
```
lemonlatte:
```suggestion
    function burnAndMint(uint256[] memory _burnedEditions) external {
```
Let's use `external` if it would not be called internally.

contracts/MoMABurn.sol
```diff
@@ -0,0 +1,136 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+import "./FeralfileArtworkV3_3.sol";
```
lemonlatte: I think we can just include the function we call so we do not need to include the whole contract.
This would save some cost for the deployment. For example,
```
contract FeralfileExhibitionV3_3 {
    function mintArtworkEdition(uint256 _artworkID, address _owner) public {}
}
```

Same for `ERC721`. We only calls few functions.

contracts/MoMABurn.sol
```diff
@@ -0,0 +1,136 @@
...
+        address _tier2Address,
+        address _tier3Address
+    ) public onlyOwner {
+        require(
+            _tier1Address != address(0) &&
+                _tier2Address != address(0) &&
+                _tier3Address != address(0),
+            "Invalid contract address"
+        );
+        tier1Contract = FeralfileExhibitionV3_3(_tier1Address);
```
lemonlatte: somehow, I feel we can only save addresses. And only make the instances when we call to it.

contracts/MoMABurn.sol
```diff
@@ -0,0 +1,136 @@
...
+    FeralfileExhibitionV3_3 public tier3Contract; // Address of new token for Tier 3
+
+    uint256 private tier1ArtworkID; // Artwork ID of Tier 1
+    uint256 private tier2ArtworkID; // Artwork ID of Tier 2
+    uint256 private tier3ArtworkID; // Artwork ID of Tier 3
+
+    bool public isBurnEnabled;
+
+    constructor() {
+        isBurnEnabled = false;
```
lemonlatte, line 60: what about set the 
- burnContractAddress
- tier1Address
- tier2Address
- tier3Address
- tier1ArtworkID
- tier2ArtworkID
- tier3ArtworkID
on deployment?

contracts/MoMABurn.sol
```diff
@@ -0,0 +1,136 @@
...
+                        address(this),
+                    "Token need to approve for this contract"
+                );
+            }
+
+            burnContract.safeTransferFrom(
+                _msgSender(),
+                DEAD_ADDRESS,
+                _burnedEditions[i]
+            );
```
lemonlatte: in here, I think it checks all for approve above. I think we can just call to this function directly.

lemonlatte (changes requested):
I feel the set function for burnContract, mintContract and artwork should be all or nothing. 
I also feel the contract is not re-useable so we do not need to update.

What do you think?

migrations/205_mint_and_burn.js
tienngovan: Updated use that variable

contracts/MoMABurn.sol
tienngovan: Updated

contracts/MoMABurn.sol
tienngovan: Updated

contracts/MoMABurn.sol
tienngovan: Updated

contracts/MoMABurn.sol
tienngovan: Added those params to constructor

contracts/MoMABurn.sol
tienngovan: Updated

contracts/MoMABurn.sol
tienngovan: Updated with MoMA artwork validation

contracts/MoMABurn.sol
```diff
@@ -0,0 +1,179 @@
...
+                    revert("Invalid burning editions");
+                }
+            }
+
+            require(
+                artwork.editionSize > 0 &&
+                    _burnedEditions[i] > burnArtworkID &&
+                    _burnedEditions[i] < burnArtworkID + artwork.editionSize,
+                "Edition ID is not support for burning"
+            );
```
lemonlatte:
```suggestion
            require(
                artwork.editionSize > 0 &&
                    _burnedEditions[i] >= burnArtworkID &&
                    _burnedEditions[i] <= burnArtworkID + artwork.editionSize,
                "Edition ID is not support for burning"
            );
```





According to FFV2.
>
     /// @notice The range of editionNumber should be between 0 (AP) ~ artwork.editionSize
     require(
          editionNumber <= artworks[artworkID].editionSize,
          "edition number exceed the edition size of the artwork"
      );

The editionID would be [ artworkID, artworkID + editionSize ].

lemonlatte (changes requested):

contracts/MoMABurn.sol
tienngovan: Updated

lpopo0856:
Just try to ask the GPT-3 is there's any vulnerability

Upon reviewing the provided smart contract, I have identified a few potential vulnerabilities:

1. In the updateArtworkCIDs function, there is no input validation to ensure that only authorized parties can call this function. It uses a modifier onlyAuthorized, but it is not defined in the contract code. This leaves the contract open to unauthorized access and could lead to loss of funds or denial of service attacks.

2. The batchMint function is defined as external pure override. As it is marked as pure, it should not modify any state variables in the contract. However, this function is not implemented, which could potentially cause issues if it is called by mistake.

3. The mintArtworkEdition function does not have input validation to ensure that the _owner parameter is a valid Ethereum address. This could lead to loss of funds if an invalid address is provided.

4. The artworkCIDs mapping is not private and can be read by anyone. This may not be a vulnerability in and of itself, but it could expose information that the contract owner did not intend to be public.

5. The baseURI variable is initialized to "ipfs://" if _tokenBaseURI is an empty string. This could potentially cause issues if the intended baseURI is not "ipfs://", and the contract owner forgets to set _tokenBaseURI before deploying the contract.

Overall, the contract appears to be relatively secure, but the above issues should be addressed to reduce potential attack vectors.

---
I think the third point really catch it!!

lpopo0856:
@tienngovan @lemonlatte We should block the burnAndMint request from contract, and make sure only EOA calling it

contracts/MoMABurn.sol
```diff
@@ -0,0 +1,193 @@
...
+        require(
+            _burnContractAddress != address(0) &&
+                _tier1Address != address(0) &&
+                _tier2Address != address(0) &&
+                _tier3Address != address(0),
+            "Invalid contract address"
+        );
+        (, , string memory burnFingerprint, uint256 burnEditionSize) = FFV2(
+            _burnContractAddress
+        ).artworks(_burnArtworkID);
```
lemonlatte: Why not declare to variable `burnArtwork` directly? It makes the code hard to read. Do we use this way because we want to save the cost?

contracts/MoMABurn.sol
```diff
@@ -0,0 +1,193 @@
...
+        require(
+            burnEditionSize > 0 &&
+                bytes(burnFingerprint).length != 0 &&
+                tier1EditionSize > 0 &&
+                bytes(tier1Fingerprint).length != 0 &&
+                tier2EditionSize > 0 &&
+                bytes(tier2Fingerprint).length != 0 &&
+                tier3EditionSize > 0 &&
+                bytes(tier3Fingerprint).length != 0,
+            "Invalid artwork ID"
```
lemonlatte:
```suggestion
            "Invalid minting parameters"
```

lemonlatte (changes requested):

contracts/MoMABurn.sol
tienngovan: Fixed

contracts/MoMABurn.sol
lemonlatte: let's create a private function for v2 and v3. For example, 
`function getV2ContractArtwork(address, artworkID) returns (uint256, string)`

So this ugly thing would only in a place.

lemonlatte (changes requested):

lemonlatte (approved):
Looks good to me.

PR #16, opened 2023-03-09
Implemented FeralFileArtwork V3_3 supporting MoMA burn

tienngovan:


contracts/FeralfileArtworkV3_3.sol
```diff
@@ -0,0 +1,133 @@
...
+        for (uint256 i = 0; i < _artworkIDs.length; i++) {
+            require(
+                bytes(artworks[_artworkIDs[i]].title).length != 0,
+                "Artwork does not exist"
+            );
+            artworkCIDs[_artworkIDs[i]] = _artworkCIDs[i];
+        }
+    }
+
+    /// @notice mintArtworkEdition use for mint token from burn contract
```
lpopo0856: `from burn contract`
remove the burn related words in v3_3 contract cause it doesn't required the minter has to be a burn contract

lpopo0856 (changes requested):
Two missing logic I found:
~~1. we miss the logic that the Tier 3 artwork could only mint 100 edition in total, we should set up a max edition mint amount in contract~~
2. we should disable mint function, so admin cant mint without burning edition. Also if admin mint through other function it'll break the counter

contracts/FeralfileArtworkV3_3.sol
```diff
@@ -0,0 +1,133 @@
...
+            bytes(artworks[_artworkID].title).length != 0,
+            "Artwork does not exist"
+        );
+
+        require(
+            bytes(artworkCIDs[_artworkID]).length != 0,
+            "Artwork CID does not exist"
+        );
+
+        tokenCounter.increment(); // Should be start from 1
```
lemonlatte: I am not sure if this is right. If there is no AE, PP for an artwork in FF now. Do we start from `1`?

contracts/FeralfileArtworkV3_3.sol
```diff
@@ -0,0 +1,133 @@
...
+        return allArtworkEditionsIndex[_tokenId].artworkID;
+    }
+
+    /// @notice updateArtworkCIDs use for update CIDs for artworks
+    /// @param _artworkIDs - array of artwork IDs
+    /// @param _artworkCIDs - array of ipfs CIDs
+    function updateArtworkCIDs(
+        uint256[] memory _artworkIDs,
+        string[] memory _artworkCIDs
+    ) public onlyOwner {
```
lemonlatte: I think we can allow authorized people to update the artworkCIDs.

test/feralfile_exhibition_v3_3.js
```diff
@@ -0,0 +1,230 @@
...
+    this.exhibition = await FeralfileExhibitionV3_3.new(
+      'Feral File V3_4 Test 001',
+      'FFV3',
+      1000,
+      '0x8fd310de32848798eB64Bd88f9C5656Eea32415e',
+      'https://ipfs.bitmark.com/ipfs/QmaptARVxNSP36PQai5oiCPqbrATvpydcJ8SPx6T6Yp1CZ',
+      IPFS_GATEWAY_PREFIX,
+      true,
+      true,
+      accounts[0]
```
lemonlatte: use owner as minter might not test the minter correctly. We can not distinguish an operation is allowed by minter or by owner.

lemonlatte (changes requested):
Need test cases:
1. mint failed with authorized addresses
2. updateArtworkCIDs failed by minter, ok for owner

contracts/FeralfileArtworkV3_3.sol
lemonlatte: > Two missing logic I found:
> 
> 1. we miss the logic that the Tier 3 artwork could only mint 100 edition in total, we should set up a max edition mint amount in contract

Doesn't this solve via creating artwork with specific `editionSize` assign? @lpopo0856

lpopo0856: Yes you're right :{ 
Sorry, I just remove the comment, but we have to make sure we put the AE, PP amount into 0

contracts/FeralfileArtworkV3_3.sol
tienngovan: Fixed start from zero

contracts/FeralfileArtworkV3_3.sol
tienngovan: Updated

test/feralfile_exhibition_v3_3.js
tienngovan: Updated

lemonlatte:
@tienngovan can you remove the formatting change in the PR?
@lpopo0856 Any thought to make the formatter using the same rule in solidity?

lpopo0856:
> @tienngovan can you remove the formatting change in the PR? @lpopo0856 Any thought to make the formatter using the same rule in solidity?

What formatter you're using? @tienngovan

lpopo0856:
Just try to ask the GPT-3.5 is there's any vulnerability

Upon reviewing the provided smart contract, I have identified a few potential vulnerabilities:

1. In the updateArtworkCIDs function, there is no input validation to ensure that only authorized parties can call this function. It uses a modifier onlyAuthorized, but it is not defined in the contract code. This leaves the contract open to unauthorized access and could lead to loss of funds or denial of service attacks.

2. The batchMint function is defined as external pure override. As it is marked as pure, it should not modify any state variables in the contract. However, this function is not implemented, which could potentially cause issues if it is called by mistake.

3. The mintArtworkEdition function does not have input validation to ensure that the _owner parameter is a valid Ethereum address. This could lead to loss of funds if an invalid address is provided.

4. The artworkCIDs mapping is not private and can be read by anyone. This may not be a vulnerability in and of itself, but it could expose information that the contract owner did not intend to be public.

5. The baseURI variable is initialized to "ipfs://" if _tokenBaseURI is an empty string. This could potentially cause issues if the intended baseURI is not "ipfs://", and the contract owner forgets to set _tokenBaseURI before deploying the contract.

Overall, the contract appears to be relatively secure, but the above issues should be addressed to reduce potential attack vectors.

lpopo0856 (approved):

contracts/FeralfileArtworkV3_3.sol
```diff
@@ -0,0 +1,140 @@
...
+        for (uint256 i = 0; i < _artworkIDs.length; i++) {
+            require(
+                bytes(artworks[_artworkIDs[i]].title).length != 0,
+                "Artwork does not exist"
+            );
+            artworkCIDs[_artworkIDs[i]] = _artworkCIDs[i];
+        }
+    }
+
+    function batchMint(
```
lemonlatte, line 106: // I think we need docs even this is a dummy function. By saying why we need this.

lemonlatte (approved):
I just leave some comments.

contracts/FeralfileArtworkV3_3.sol
tienngovan: Added comments

PR #19, opened 2023-06-05
Feralfile Exhibition V4

tienngovan:


contracts/FeralfileArtworkV4.sol
```diff
@@ -90,23 +158,144 @@ contract FeralfileExhibitionV4 is
...
+        for (uint256 i = 0; i < tokenIds_.length; i++) {
+            _safeTransfer(address(this), destination_, tokenIds_[i], "");
+
+            emit BuyArtwork(destination_, tokenIds_[i]);
+        }
+        // distribute royalty
+        for (uint256 i = 0; i < saleData_.royalties.length; i++) {
+            payable(saleData_.royalties[i].recipient).transfer(
+                (msg.value * saleData_.royalties[i].bps) / 10000
+            );
```
lemonlatte: I think this can more into the previous for loop

contracts/FeralfileArtworkV4.sol
```diff
@@ -90,23 +158,144 @@ contract FeralfileExhibitionV4 is
...
+        );
+        require(
+            saleData_.expiryTime > block.timestamp,
+            "FeralfileExhibitionV4: sale is expired"
+        );
+        require(
+            saleData_.cost == msg.value,
+            "FeralfileExhibitionV4: invalid payment amount"
+        );
+
```
lemonlatte, line 532: validate the length of `saleData_.royalties` is identical to `tokenIds_`

contracts/FeralfileArtworkV4.sol
```diff
@@ -90,23 +158,144 @@ contract FeralfileExhibitionV4 is
...
+    /// @param artworkIndex_ - the index of artwork in series
+    /// @param ipfsCID_ - the ipfs cid of artwork
+    function _mintArtwork(
+        uint256 seriesIndex_,
+        uint256 artworkIndex_,
+        string memory ipfsCID_
+    ) internal {
+        require(canMint, "FeralfileExhibitionV4: not in minting stage");
+        require(
+            seriesIndex_ >= 0,
```
lemonlatte: Redundant, `uint256` is always >= 0

contracts/FeralfileArtworkV4.sol
```diff
@@ -90,23 +158,144 @@ contract FeralfileExhibitionV4 is
...
+            "FeralfileExhibitionV4: invalid series index"
+        );
+
+        require(
+            !registeredIPFSCIDs[ipfsCID_],
+            "FeralfileExhibitionV4: IPFS cid already registered"
+        );
+
+        uint256 artworkID = (seriesIndex_ + 1) *
```
lemonlatte: I think we do not need to plus more. I think It is better leave this logic to FF server.

lemonlatte (changes requested):

contracts/FeralfileArtworkV4.sol
tienngovan: Fixed

contracts/FeralfileArtworkV4.sol
tienngovan: Fixed

contracts/FeralfileArtworkV4.sol
tienngovan: added

contracts/FeralfileArtworkV4.sol
tienngovan: Fixed

contracts/FeralfileArtworkV4.sol
```diff
@@ -1,112 +1,602 @@
...
+        // emit event
+        emit NewArtwork(owner, seriesId, tokenId);
+    }
+
+    /// @notice Burn a collection of artworks
+    /// @dev the function iterates over the array of token ID to call the internal function _burnArtwork
+    /// @param tokenIds an array of token ID
+    function burnArtworks(uint256[] memory tokenIds) external {
+        for (uint256 i = 0; i < tokenIds.length; i++) {
+            _burnArtwork(tokenIds[i]);
```
lemonlatte, line 573:
```suggestion
            require(_isApprovedOrOwner(_msgSender(), tokenId), "ERC721: caller is not token owner or approved");
            _burnArtwork(tokenIds[i]);
```
@jollyjoker992 the external function can be called for users so we need to check the ownership before burn tokens.

lemonlatte (changes requested):

contracts/FeralfileArtworkV4.sol
```diff
@@ -1,112 +1,602 @@
...
     }
 
-    function supportsInterface(
-        bytes4 interfaceId
-    ) public view virtual override(ERC721Enumerable, IERC165) returns (bool) {
-        return
-            interfaceId == type(IERC721Enumerable).interfaceId ||
-            super.supportsInterface(interfaceId);
+    /// @notice Stop token selling and transfer remaining tokens back to the underlying addresses
+    function stopSaleAndTransfer(
```
lemonlatte, line 260: There is a chance to stop the sale with only part of series transfer tokens to addresses. This means the sale can start with the remaining tokens again. I think we need to ensure either:
- the length of `seriesIds` is same to the length of `_seriesMaxSupplies`
- balanceOf(address(this)) == 0

---

- The first approach would need to introduce another key since we can not count the length of a map directly.
- The later one would waste more gas since we can only check after we burn all tokens.

lemonlatte (changes requested):
@jollyjoker992

lemonlatte (dismissed):
LGTM

contracts/FeralfileArtworkV4.sol
```diff
@@ -1,112 +1,614 @@
...
-        bool isBurnable_,
-        bool isBridgeable_
-    ) ERC721(name_, symbol_) {
-        isBurnable = isBurnable_;
-        isBridgeable = isBridgeable_;
+        address signer_,
+        bool burnable_,
+        bool bridgeable_,
+        address vault_,
+        address costReceiver_,
```
lpopo0856: Small: put the same type params in order. though it wont save gas in function param, but it's better for reading

contracts/FeralfileArtworkV4.sol
```diff
@@ -1,112 +1,614 @@
...
+    /// @return Artwork the Artwork object
+    function getArtwork(
+        uint256 tokenId
+    ) external view virtual returns (Artwork memory) {
+        require(_exists(tokenId), "ERC721: invalid token ID");
+        return _allArtworks[tokenId];
+    }
+
+    /// @notice Set vault contract
+    /// @dev don't allow to set signer as zero address
```
lpopo0856: typo: set vault

contracts/FeralfileArtworkV4.sol
```diff
@@ -1,112 +1,614 @@
...
+        );
+    }
+
+    /// @notice Start token selling
+    function startSale() external onlyOwner {
+        mintable = false;
+        resumeSale();
+    }
+
+    /// @notice Resume token selling
```
lpopo0856: small: consistent wording between `selling` and `sale`

contracts/FeralfileArtworkV4.sol
```diff
@@ -1,112 +1,614 @@
...
+            "FeralfileExhibitionV4: mintable required to be false"
+        );
+        require(
+            _selling,
+            "FeralfileExhibitionV4: _selling required to be true"
+        );
+        _selling = false;
+    }
+
+    /// @notice Stop token selling and burn remaining tokens
```
lpopo0856: same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/19/files#r1243108317)

contracts/FeralfileArtworkV4.sol
```diff
@@ -1,112 +1,614 @@
...
+        pauseSale();
+
+        // burn remaining tokens
+        uint256[] memory tokenIds = _ownedTokens[address(this)];
+        for (uint256 i = 0; i < tokenIds.length; i++) {
+            _burnArtwork(tokenIds[i]);
+        }
+    }
+
+    /// @notice Stop token selling and transfer remaining tokens back to the underlying addresses
```
lpopo0856, line 259: same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/19/files#r1243108317)

contracts/FeralfileArtworkV4.sol
```diff
@@ -1,112 +1,614 @@
...
+        tokenBaseURI = baseURI_;
     }
 
-    /// @notice A URL for the opensea storefront-level metadata
-    function contractURI() public view returns (string memory) {
-        return _contractURI;
+    /// @notice the cost receiver address
+    /// @param costReceiver_ - the address of cost receiver
+    function setCostReceiver(address costReceiver_) external onlyOwner {
+        costReceiver = costReceiver_;
```
lpopo0856, line 384: missing `costReceiver_ != address(0)` check

contracts/FeralfileArtworkV4.sol
```diff
@@ -1,112 +1,614 @@
...
+        );
+
+        return string(abi.encodePacked(tokenBaseURI, "/", tokenId.toString()));
     }
 
     /// @notice Update the base URI for all tokens
-    function setTokenBaseURI(string memory baseURI_) external onlyAuthorized {
-        _tokenBaseURI = baseURI_;
+    function setTokenBaseURI(string memory baseURI_) external onlyOwner {
+        tokenBaseURI = baseURI_;
```
lpopo0856, line 374: missing `bytes(baseURI_).length > 0` check

contracts/FeralfileArtworkV4.sol
```diff
@@ -1,112 +1,614 @@
...
+        address from = address(this);
+        uint256[] memory tokenIds = _ownedTokens[from];
+        for (uint256 i = 0; i < tokenIds.length; i++) {
+            uint256 tokenId = tokenIds[i];
+            Artwork memory artwork = _allArtworks[tokenId];
+
+            for (uint16 j = 0; j < seriesIds.length; j++) {
+                if (artwork.seriesId == seriesIds[j]) {
+                    address to = recipientAddresses[j];
+                    _transfer(from, to, tokenId);
```
lpopo0856: why not using `_safeTransfer`?

contracts/FeralfileArtworkV4.sol
jollyjoker992: @lpopo0856 I dont think so, we should put the params based on the meaning like something is similar could put together. In this case, i will bring the signer next to vault and cost receiver.

lpopo0856: ok!

lpopo0856 (dismissed):

jollyjoker992 (dismissed):

lemonlatte (approved):
LGTM

PR #20, opened 2023-06-06
Remove ipfs and add vault contract

lpopo0856:
As title.
- Remove the ipfs required when minting, and customized the tokenURI return to compose a directory url.
- Add vault contract support and modify related buyArtwork logic
- Add more signing argument to make it safe and adjust testing
- Redefine the `saleData.cost`

### DIFF
--- a/contracts/FeralfileArtworkV4.sol
+++ b/contracts/FeralfileArtworkV4.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.13;
 
 import "@openzeppelin/contracts/token/ERC721/extensions/ERC721Enumerable.sol";
 import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
@@ -8,7 +8,10 @@ import "@openzeppelin/contracts/utils/Strings.sol";
 
 import "./Authorizable.sol";
 import "./UpdateableOperatorFilterer.sol";
-import "./FeralfileVault.sol";
+
+contract Vault {
+    function pay(uint256 weiAmount) external {}
+}
 
 contract FeralfileExhibitionV4 is
     IERC165,
@@ -27,6 +30,9 @@ contract FeralfileExhibitionV4 is
     // signer of buy artwork
     address public signer = 0x0000000000000000000000000000000000000000;
 
+    // vault contract
+    address public vault = 0x0000000000000000000000000000000000000000;
+
     // burnable
     bool public isBurnable;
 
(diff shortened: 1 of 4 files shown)


tienngovan (approved):

contracts/FeralfileArtworkV4.sol
```diff
@@ -118,11 +112,10 @@ contract FeralfileExhibitionV4 is
...
         string memory baseURI = _tokenBaseURI;
-        if (bytes(baseURI).length == 0) {
-            baseURI = "ipfs://";
-        }
-
-        return string(abi.encodePacked(baseURI, artworks[tokenId_].ipfsCID));
+        return
+            bytes(baseURI).length > 0
+                ? string.concat(baseURI, "/", Strings.toString(tokenId_))
+                : "";
```
lemonlatte, line 123: I think we should return error not return empty

contracts/FeralfileArtworkV4.sol
```diff
@@ -272,7 +262,7 @@ contract FeralfileExhibitionV4 is
             // distribute royalty
             for (uint256 j = 0; j < saleData_.royalties[i].length; j++) {
                 payable(saleData_.royalties[i][j].recipient).transfer(
-                    (itemCost * saleData_.royalties[i][j].bps) / 10000
+                    (itemRevenue * saleData_.royalties[i][j].bps) / 10000
```
lemonlatte: just aware that we should use something like `revenueShares` instead of `royalties`.

contracts/FeralfileArtworkV4.sol
```diff
@@ -241,24 +231,24 @@ contract FeralfileExhibitionV4 is
...
+            abi.encode(block.chainid, address(this), saleData_)
+        );
 
         require(
             isValidRequest(requestHash, signer, r_, s_, v_),
             "FeralfileExhibitionV4: invalid signature"
         );
 
-        uint256 itemCost = saleData_.cost / saleData_.tokenIds.length;
+        uint256 itemRevenue = (saleData_.price - saleData_.cost) / saleData_.tokenIds.length;
```
lemonlatte: I think it is better to use SafeMath to deal with fund spliting

lemonlatte (changes requested):

tienngovan (approved):

contracts/FeralfileArtworkV4.sol
lpopo0856: I'll fix this after HIeu's pr merged after this PR

contracts/FeralfileArtworkV4.sol
lpopo0856: no need for safemath after 0.8

PR #21, opened 2023-06-06
Mint functions

jollyjoker992:


contracts/FeralfileArtworkV4.sol
```diff
@@ -10,60 +10,216 @@ import "./Authorizable.sol";
...
+        return super.supportsInterface(interfaceId);
     }
 
-    /// @notice Return a count of series registered in this exhibition
-    function totalSeries() public view virtual returns (uint256) {}
+    function mintable() public view virtual returns (bool) {
+        return _mintable;
+    }
+
+    function setMintable(bool mintable_) external virtual onlyAuthorized {
```
lemonlatte: We should not make 
- setMintable
- setBurnable
- setBridgeable

be changed directly. It means FF can mint anything anytime if we want.
For open edition, I plan not do it over the new contract. And just make this a pure exhibition contract. So I think we only need to disable it when a sale starts and leave it false forever.

contracts/FeralfileArtworkV4.sol
```diff
@@ -76,37 +232,103 @@ contract FeralfileExhibitionV4 is
...
+        require(
+            _seriesTotalSupplies[seriesId] < _seriesMaxSupplies[seriesId],
+            "FeralfileExhibitionV4: no slots available"
+        );
+
+        // mint
+        _totalSupply += 1;
+        _seriesTotalSupplies[seriesId] += 1;
+        _allArtworks[tokenId] = Artwork(seriesId, tokenId);
+        _safeMint(owner, tokenId);
```
lemonlatte: - with _safeMint, we need `IERC721Receiver` for this contract to receive tokens.

I think we can just use `_mint`, if we know what we are doing

lemonlatte (changes requested):

contracts/FeralfileArtworkV4.sol
```diff
@@ -10,60 +10,216 @@ import "./Authorizable.sol";
...
+            "FeralfileExhibitionV4: signer_ is zero address"
+        );
+        require(
+            seriesIds_.length > 0,
+            "FeralfileExhibitionV4: seriesIds_ is empty"
+        );
+        require(
+            seriesMaxSupplies_.length > 0,
+            "FeralfileExhibitionV4: _seriesMaxSupplies is empty"
+        );
```
lpopo0856, line 135: 1/1 is not a series, should we move that rule to contract level?
just bring up the discussion, I'm ok to put a series for 1/1 on contract/FF DB level but hiding from UI :)

contracts/FeralfileArtworkV4.sol
```diff
@@ -10,60 +10,216 @@ import "./Authorizable.sol";
...
 
     function supportsInterface(
         bytes4 interfaceId
-    ) public view virtual override(ERC721Enumerable, IERC165) returns (bool) {
-        return
-            interfaceId == type(IERC721Enumerable).interfaceId ||
-            super.supportsInterface(interfaceId);
+    ) public view virtual override(ERC721) returns (bool) {
+        return super.supportsInterface(interfaceId);
     }
```
lpopo0856: We could remove this override, cause now we only have ERC721 that implement this function

contracts/FeralfileArtworkV4.sol
```diff
@@ -10,60 +10,216 @@ import "./Authorizable.sol";
...
-    /// @notice Return a count of series registered in this exhibition
-    function totalSeries() public view virtual returns (uint256) {}
+    function mintable() public view virtual returns (bool) {
+        return _mintable;
+    }
+
+    function setMintable(bool mintable_) external virtual onlyAuthorized {
+        _mintable = mintable_;
+    }
+
```
lpopo0856: We could remove the not used virtual in following functions, I don't think we're planing to override these functions in future.

contracts/FeralfileArtworkV4.sol
```diff
@@ -10,60 +10,216 @@ import "./Authorizable.sol";
...
+    function burnable() public view virtual returns (bool) {
+        return _burnable;
+    }
```
lpopo0856: functions like this are redundant, we could just set the `_burnable` to public so it could be read directly

contracts/FeralfileArtworkV4.sol
```diff
@@ -10,60 +10,216 @@ import "./Authorizable.sol";
...
+        require(_exists(tokenId), "ERC721: token doesn't exist");
+        return _allArtworks[tokenId];
+    }
+
+    function setApprovalForAll(
+        address operator,
+        bool approved
+    ) public override(ERC721) onlyAllowedOperatorApproval(operator) {
+        super.setApprovalForAll(operator, approved);
+    }
```
lpopo0856, line 222: remove this. we no need to override cause you don't add any customized logic.
and ERC721 is the only contract that implement the `setApprovalForAll`

contracts/FeralfileArtworkV4.sol
```diff
@@ -10,60 +10,216 @@ import "./Authorizable.sol";
...
+        super.approve(operator, tokenId);
+    }
+
+    function transferFrom(
+        address from,
+        address to,
+        uint256 tokenId
+    ) public override(ERC721) onlyAllowedOperator(from) {
+        super.transferFrom(from, to, tokenId);
+    }
```
lpopo0856, line 189: same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/21/files#r1220690401)

contracts/FeralfileArtworkV4.sol
```diff
@@ -10,60 +10,216 @@ import "./Authorizable.sol";
...
+
+    function safeTransferFrom(
+        address from,
+        address to,
+        uint256 tokenId
+    ) public override(ERC721) onlyAllowedOperator(from) {
+        super.safeTransferFrom(from, to, tokenId);
+    }
```
lpopo0856, line 197: same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/21/files#r1220690401)

contracts/FeralfileArtworkV4.sol
```diff
@@ -10,60 +10,216 @@ import "./Authorizable.sol";
...
+    function safeTransferFrom(
+        address from,
+        address to,
         uint256 tokenId,
-        string memory ipfsCID
-    ) external onlyAuthorized {}
+        bytes memory data
+    ) public override(ERC721) onlyAllowedOperator(from) {
+        super.safeTransferFrom(from, to, tokenId, data);
+    }
```
lpopo0856, line 206: same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/21/files#r1220690401)

contracts/FeralfileArtworkV4.sol
```diff
@@ -10,60 +10,216 @@ import "./Authorizable.sol";
...
+            _exists(tokenId),
+            "ERC721Metadata: URI query for nonexistent token"
+        );
+
+        string memory baseURI = _tokenBaseURI;
+        if (bytes(baseURI).length == 0) {
+            baseURI = "ipfs://";
+        }
+
+        return string(abi.encodePacked(baseURI, "/", tokenId.toString()));
```
lpopo0856: if the `baseURI == ""` then your return will be `ipfs://tokenID` which's not correct. we should return empty when we find the `baseURI == ""`, that mean the metadata is not ready yet

contracts/FeralfileArtworkV4.sol
```diff
@@ -76,37 +232,103 @@ contract FeralfileExhibitionV4 is
...
-            r_,
-            s_
+        return
+            ECDSA.recover(ECDSA.toEthSignedMessageHash(message_), v_, r_, s_) ==
+            owner_;
+    }
+
+    function _seriesExists(uint256 seriesId) internal view returns (bool) {
+        return _seriesMaxSupplies[seriesId] > 0;
+    }
```
lpopo0856, line 394: this logic could be a modifier so we could reuse it cleaner

contracts/FeralfileArtworkV4.sol
```diff
@@ -76,37 +232,103 @@ contract FeralfileExhibitionV4 is
...
+    function _burnArtwork(uint256 tokenId) internal {
+        require(burnable(), "FeralfileExhibitionV4: token is not burnable");
+        require(_exists(tokenId), "ERC721: token doesn't exist");
 
+        // burn artwork
+        Artwork memory artwork = _allArtworks[tokenId];
+        _seriesTotalSupplies[artwork.seriesId] -= 1;
+        _totalSupply -= 1;
+        _burn(tokenId);
+
```
lpopo0856, line 460: are we missing to remove tokenID from  `_allArtworks` ?

.gitignore
```diff
@@ -3,3 +3,4 @@ build
 node_modules
 coverage/*
 coverage.json
+.vscode
```
lpopo0856: we should keep this, it include the setting of prettier

lpopo0856 (changes requested):
some review

contracts/FeralfileArtworkV4.sol
lemonlatte: this is for `operatorRegistry` so it is required.

contracts/FeralfileArtworkV4.sol
lemonlatte: I do not think so!! if so, why not turn `_exists` into a modifier?

contracts/FeralfileArtworkV4.sol
jollyjoker992: @lemonlatte I just want to make it more flexible. If we don't concern about the security risks, why dont we keep them?

contracts/FeralfileArtworkV4.sol
jollyjoker992: @lpopo0856 How can you deal with 1/1 without series?

contracts/FeralfileArtworkV4.sol
jollyjoker992: @lpopo0856 I'm not sure about the best practice for smart contract but for normal application code, we shouldn't expose variable publicly rather than using getter/setter. Can you explain a little bit more your idea?

contracts/FeralfileArtworkV4.sol
jollyjoker992: I think it depends how to use it. modifier is quite strict that can only be run before executing the target function while a private function could be called flexibly.

contracts/FeralfileArtworkV4.sol
lemonlatte: When people look into the contract, they know that we CAN do this. If the policy is not do, it would be better we let people see this from the contract instead of leave some fuzzy space for us to do in the future.

contracts/FeralfileArtworkV4.sol
lpopo0856: there's no difference to put variables public and creating a separate function to get it. I prefer just expose them

contracts/FeralfileArtworkV4.sol
lpopo0856: Oh I see, I'll remove the related comments

contracts/FeralfileArtworkV4.sol
lpopo0856: hmm I do some research and seems like the best practice is using the modifier so it could be reuse and override.
but actually using private function saving more gas while deploy the contract cause modifier copy paste the code into the function that use it while compiling.

lpopo0856: I'll change my mind and vote for private function :P

contracts/FeralfileArtworkV4.sol
lpopo0856: solidity compiler will create a getter function automatically for public variable

PR #23, opened 2023-06-07
Signature validation for vault

lpopo0856:
As title, change vault to use signature validation for `payForSale` request.

- Refactor the duplicated component into separate contracts
  - `ECDSASign` to handle the signature related function
  - `FeralfileSaleStruct` to record the saleData structure


contracts/ECDSASign.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+contract ECDSASign is Ownable {
+    address private signer;
```
jollyjoker992: Minor. Should use underscore for private/internal.

contracts/FeralfileSaleStruct.sol
```diff
@@ -0,0 +1,34 @@
...
+        uint256 price; // in wei
+        uint256 cost; // in wei
+        uint256 expiryTime;
+        address destination;
+        uint256[] tokenIds;
+        RevenueShare[][] revenueShares; // address and royalty bps (500 means 5%)
+        bool payByVaultContract; // get eth from vault contract, used by credit card pay that proxy by ITX
+    }
+
+    function isValidSaleData(SaleData calldata saleData_) internal view {
```
lemonlatte:
```suggestion
    function validateSaleData(SaleData calldata saleData_) internal view {
```

contracts/ECDSASign.sol
```diff
@@ -0,0 +1,46 @@
...
+        _signer = signer_;
+    }
+
+    /// @notice verifySignature validates a message by ecrecover to ensure
+    //          it is signed by signer.
+    /// @param message_ - the raw message for signing
+    /// @param r_ - part of signature for validating parameters integrity
+    /// @param s_ - part of signature for validating parameters integrity
+    /// @param v_ - part of signature for validating parameters integrity
+    function verifySignature(
```
lemonlatte:
```suggestion
    function isValidSignature(
```

contracts/FeralfileVault.sol
```diff
@@ -3,34 +3,51 @@ pragma solidity ^0.8.0;
...
+        );
+        require(!paidSale[requestHash], "FeralfileVault: paid sale");
+        require(
+            verifySignature(requestHash, r_, s_, v_),
+            "FeralfileVault: invalid signature"
+        );
+        require(
+            address(this).balance >= saleData_.price,
+            "FeralfileVault: insufficient balance"
+        );
```
lemonlatte: we can validate this at beginning

contracts/FeralfileSaleStruct.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+contract FeralfileSaleStruct {
```
lemonlatte:
```suggestion
contract FeralfileSaleData {
```

contracts/FeralfileArtworkV4.sol
```diff
@@ -141,9 +118,8 @@ contract FeralfileExhibitionV4 is
         burnable = burnable_;
         bridgeable = bridgeable_;
         costReceiver = costReceiver_;
-        vault = vault_;
+        vault = FeralfileVault(payable(vault_));
```
lemonlatte: why not just the address and need to make the whole instance as a public variable?

contracts/FeralfileVault.sol
```diff
@@ -3,34 +3,51 @@ pragma solidity ^0.8.0;
...
-    }
-
-    function unsetExhibitionContract(address ec) external onlyOwner {
-        exhibitionContract[ec] = false;
-    }
-
-    function pay(uint256 amount) external onlyExhibitionContract {
-        require(address(this).balance >= amount, "insufficient balance");
-        payable(msg.sender).transfer(amount);
+        isValidSaleData(saleData_);
```
lemonlatte: require(msg.sender == form) to prevent other people to run this.

lemonlatte (changes requested):

contracts/ECDSASign.sol
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+contract ECDSASign is Ownable {
```
lemonlatte:
```suggestion
contract FeralFileSigner is Ownable {
```
or 
```suggestion
contract ECDSASigner is Ownable {
```

lemonlatte (changes requested):

contracts/FeralfileArtworkV4.sol
lpopo0856: I'm thinking that init a instance everytime in buyArtwork seems wasted. we could do this once while contract init.
But I don't do the test if the init instance cost much, maybe it doesn't cost any :P
my thought is it just no harm to put it at constructor level.

contracts/ECDSASign.sol
lpopo0856: I'll pick ECDSASigner

contracts/ECDSASign.sol
lpopo0856: I take this name form Hieu. we change this naming back and force. @jollyjoker992 do you have strong opinion here?

lemonlatte: I think `isXX` return bool. `verfifyXX` can just check and revert like `validateSaleData` above

jollyjoker992: @lpopo0856 No, i dont. `validSignature()` or `isValidSignature` could be ok. 
BTW, i used to name a function return bool without `is` but it's fine, it's up to you.

contracts/FeralfileArtworkV4.sol
lemonlatte: If so, I think we make the `vault` a private variable and create another variable in public variable just for address.

I do not know what people will see in etherscan, but if vault is an object, it looks like can not get help information when we check from the etherscan.

contracts/FeralfileArtworkV4.sol
```diff
@@ -462,6 +393,19 @@ contract FeralfileExhibitionV4 is
         emit BurnArtwork(tokenId);
     }
 
+    /// @notice able to recieve funds from vault contract
```
jollyjoker992: Minor typo.

contracts/FeralfileVault.sol
lpopo0856: I remove the from and take msg.sender as param to make it straight

contracts/FeralfileArtworkV4.sol
```diff
@@ -462,6 +393,19 @@ contract FeralfileExhibitionV4 is
...
+    receive() external payable {
+        require(
+            msg.sender == address(vault),
+            "FeralfileExhibitionV4: only accept fund from vault contract."
+        );
+    }
+
+    /// @notice withdraw all funds
+    function withdrawFunds() external onlyOwner {
```
jollyjoker992, line 416: Why don't you allow withdraw an amount instead of the whole one?

lpopo0856: this's a function that incase accident happen, we won't have fund lock in this contract forever, no need to specify amount.
the contract balance should always be zero if everything work normally.

contracts/FeralfileVault.sol
```diff
@@ -3,34 +3,51 @@ pragma solidity ^0.8.0;
...
+        bytes32 requestHash = keccak256(
```
jollyjoker992: `message` could be better.

contracts/FeralfileVault.sol
```diff
@@ -3,34 +3,51 @@ pragma solidity ^0.8.0;
...
+        );
+        require(
+            address(this).balance >= saleData_.price,
+            "FeralfileVault: insufficient balance"
+        );
+        _paidSale[requestHash] = true;
+        payable(from).transfer(saleData_.price);
     }
 
     function withdrawFunds() external onlyOwner {
```
jollyjoker992: Same [question](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/23/files#r1222569397).

jollyjoker992 (approved):

PR #28, opened 2023-10-23
Feature/1706 english auction

tienngovan:


contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,317 @@
...
+        if (!lastBid_.fromFeralFile) {
+            payable(vaultAddr_).transfer(lastBid_.amount);
+        }
+        FFV4(tokenAddr_).buyArtworks(r_, s_, v_, saleData_);
+    }
+
+    event NewBid(
+        bytes32 indexed auctionId,
+        address indexed bidder,
+        uint256 indexed amount
```
lpopo0856: could you help to emit fromferalfile too?

tienngovan: added

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+    function buyArtworks(
+        bytes32 r_,
+        bytes32 s_,
+        uint8 v_,
+        SaleData calldata saleData_
+    ) external payable {}
+}
+
+contract FeralfileEnglishAuction is Ownable, IFeralfileSaleData, ECDSASigner {
+    struct EnglishAuction {
```
jollyjoker992: Minor. It could be `Auction` is enough since the contract is `FeralFileEnglishAuction` already.

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+        bool isSettled;
+    }
+
+    struct Bid {
+        address bidder;
+        uint256 amount;
+        bool fromFeralFile;
+    }
+
+    struct LatestAuctionStatus {
```
jollyjoker992: ~I think this structure could be redundant. Instead, we could query the auction info and the highest bid separately, right?~

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+        Bid latestBid;
+        uint256 endAt;
+        bool isSettled;
+    }
+
+    mapping(bytes32 => EnglishAuction) public auctions;
+    mapping(bytes32 => Bid) public auctionLatestBid;
```
jollyjoker992: I'd suggest changing this name to `highestBids`

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+        // hash the (contractAddr + ordered tokenIDs)
+        bytes32 message = keccak256(
+            abi.encode(block.chainid, tokenContractAddr_, tokenIDs_)
+        );
+
+        // return hash as identifier
+        return message;
+    }
+
+    function auctionOngoing(bytes32 aucId_) public view returns (bool) {
```
jollyjoker992: I'd suggest changing this function name to `ongoingAuction`, the variable could be `id_` and the function modifier should be `external`.

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+        // check the auction status and return
+        require(
+            auctions[aucId_].id != 0,
+            "FeralfileEnglishAuction: auction not found"
+        );
+
+        return
```
jollyjoker992, line 58: `auctions[aucId_].endAt > block.timestamp` will be better.

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+            auctions[aucId_].startAt <= block.timestamp &&
+            auctions[aucId_].endAt >= block.timestamp;
+    }
+
+    function setEnglishAuctions(
```
jollyjoker992: I'd suggest changing this function name to `createAuctions()`

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+                "FeralfileEnglishAuction: auction start time should be in the future"
+            );
+
+            require(
+                auctions_[i].endAt > auctions_[i].startAt,
+                "FeralfileEnglishAuction: auction end time should be after start time"
+            );
+
+            require(
+                auctions_[i].extendDuration >= 0,
```
jollyjoker992: Redundant check since the `extendDuration` is `uint256`

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+                "FeralfileEnglishAuction: auction extend duration should be greater than or equal 0"
+            );
+
+            require(
+                auctions_[i].extendThreshold >= 0,
```
jollyjoker992: The same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/28/files#r1375774811)

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+            require(
+                auctions_[i].minPrice > 0,
+                "FeralfileEnglishAuction: auction min price should be greater than 0"
+            );
+
+            auctions[auctions_[i].id] = auctions_[i];
+        }
+    }
+
+    function listAuctionLatestStatus(
```
jollyjoker992: According to [this comment](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/28/files#r1375766259), i'd suggest changing this function name to `listHighestBids`

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+            "FeralfileEnglishAuction: auction not started or ended"
+        );
+        // check if auction is not settled
+        require(
+            !auction_.isSettled,
+            "FeralfileEnglishAuction: auction already settled"
+        );
+        // check if current winner not equal to current bidder (optional)
+        require(
+            lastBid_.bidder != bidder_,
```
jollyjoker992: I think we should allow them to increase his bid, why do we need to prevent?

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+            "FeralfileEnglishAuction: bidder is the current winner"
+        );
+        // check bidding amount > highest bid amount
+        require(
+            amount_ > lastBid_.amount,
```
jollyjoker992: The check based on minIncreaseAmount and minIncreaseFactor could cover this check, i think.

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+            EnglishAuction memory auction_ = auctions[aucIds_[i]];
+            // get the last auction bid object
+            results[i].latestBid = auctionLatestBid[aucIds_[i]];
+            results[i].endAt = auction_.endAt;
+            results[i].isSettled = auction_.isSettled;
+        }
+        return results;
+    }
+
+    function isValidBidRequest(
```
jollyjoker992: I'd suggest changing function name to `isValidNewBid`

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+            _isValidBidRequest(
+                auction_,
+                lastBid_,
+                bidder_,
+                amount_,
+                block.timestamp
+            );
+    }
+
+    function _isValidBidRequest(
```
jollyjoker992: I feel like this function is inconsistent. It returns a `bool` means that it validate the new bid if it's correct but somehow, we also throw error when validating the bid. I'd suggest either throw error (dont return `bool`) or return `bool` without any errors.

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+            // check bidding amount follow the minimum price
+            require(
+                amount_ >= auction_.minPrice,
+                "FeralfileEnglishAuction: bid amount should be greater than minimum price"
+            );
+        }
+        return true;
+    }
+
+    function bidOnAuction(bytes32 aucId_) external payable {
```
jollyjoker992: I'd suggest to rename function to `createBid()` and `createSignedBid()`.

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+            payable(lastBid_.bidder).transfer(lastBid_.amount);
+        }
+        // replace the new bid to bid map
+        auctionLatestBid[aucId_] = Bid({
+            bidder: msg.sender,
+            amount: msg.value,
+            fromFeralFile: false
+        });
+        // if the gap to end time is lower than threshold, update the auction end time to now() + extend duration
+        if (auction_.endAt - block.timestamp < auction_.extendThreshold) {
```
jollyjoker992: It would be `<=` .

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+        uint8 v_
+    ) external onlyOwner {
+        // transfer winning bid amount to Feral File Vault contract if winning bid is crypto bid
+        bytes32 aucId_ = getAuctionID(tokenAddr_, saleData_.tokenIds);
+
+        EnglishAuction memory auction_ = auctions[aucId_];
+        require(auction_.id != 0, "FeralfileEnglishAuction: auction not found");
+
+        require(
+            auction_.endAt < block.timestamp,
```
jollyjoker992: It should be `<=`

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,378 @@
...
+        uint256 endAt;
+        bool isSettled;
+    }
+
+    mapping(bytes32 => EnglishAuction) public auctions;
+    mapping(bytes32 => Bid) public auctionLatestBid;
+
+    constructor(address signer_) ECDSASigner(signer_) {}
+
+    function getAuctionID(
```
jollyjoker992: I believe we don't need to maintain a specific logic for the auctionID, we just need to create rule on the backend, similar to what we did for token mint. 
The reason i think is 
- Create more complexity to maintain a specific logic. We also need to think about the compatibility if we update the logic in the future. 
- Add more computation for every uses. 

Suggestion: 
Use specific algorithm like UUID->decimal to maintain the auctionID. It could be even more easier on backend side.

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,334 @@
...
+        return
+            auctions[id_].startAt <= block.timestamp &&
+            auctions[id_].endAt >= block.timestamp;
+    }
+
+    function registerAuctions(Auction[] calldata auctions_) external onlyOwner {
+        // for-loop to set the auction
+        // able to reset an auction if there's no winner and highest bid = 0
+        for (uint256 i = 0; i < auctions_.length; i++) {
+            Auction memory auc_ = auctions[auctions_[i].id];
```
jollyjoker992: Minor. Can we search all `auc` and replace with `auction`?

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,334 @@
...
+
+            require(
+                auc_.id == 0,
```
jollyjoker992, line 69: It should be `auction.id > 0`.

contracts/FeralfileEnglishAuction.sol
jollyjoker992: @tienngovan Any concern here? We. could figure a better name, i dont think `isValidBidRequest` is proper.

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    function isValidBidRequest(
+        uint256 auctionID_,
+        address bidder_,
+        uint256 amount_
+    ) external view returns (bool) {
+        // get the auction object
+        Auction memory auction_ = auctions[auctionID_];
+        // get the last auction bid object
```
jollyjoker992: Should be highest bid.

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,334 @@
...
+        }
+        return results;
+    }
+
+    function isValidBidRequest(
+        uint256 auctionID_,
+        address bidder_,
+        uint256 amount_
+    ) external view returns (bool) {
+        // get the auction object
```
jollyjoker992: I see you left some the same comments in different functions but to me, it could be redundant. Those are pretty easy to understand and the code can explain itself.

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,334 @@
...
+        Bid memory highestBid_,
+        address bidder_,
+        uint256 amount_,
+        uint256 timestamp_
+    ) internal pure {
+        // make sure the auction exist
+        require(auction_.id > 0, "FeralfileEnglishAuction: auction not found");
+        // check if auction is on going
+        require(
+            auction_.startAt <= timestamp_ && timestamp_ <= auction_.endAt,
```
jollyjoker992: Should be ` auction_.startAt <= timestamp_ && timestamp_ < auction_.endAt`

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,334 @@
...
+        if (!highestBid_.fromFeralFile) {
+            payable(highestBid_.bidder).transfer(highestBid_.amount);
+        }
+        // replace the new bid to bid map
+        highestBids[auctionID_] = Bid({
+            bidder: bidder_,
+            amount: amount_,
+            fromFeralFile: true
+        });
+        // if the gap to end time is lower than threshold, update the auction end time to now() + extend duration
```
jollyjoker992: Minor. What is `now()`? 🤔

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,334 @@
...
+            !auction_.isSettled,
+            "FeralfileEnglishAuction: auction already settled"
+        );
+        // check if current winner not equal to current bidder (optional)
+        require(
+            highestBid_.bidder != bidder_,
+            "FeralfileEnglishAuction: bidder is the current winner"
+        );
+
+        if (highestBid_.amount > 0) {
```
jollyjoker992, line 149: It should be `highestBid_.id > 0` for avoid the confuse.

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,334 @@
...
+        if (auction_.endAt - block.timestamp <= auction_.extendThreshold) {
+            auctions[auctionID_].endAt =
+                block.timestamp +
+                auction_.extendDuration;
+        }
+        // emit new bid event
+        emit NewBid(auctionID_, msg.sender, msg.value, false);
+    }
+
+    function placeSignedBid(
```
jollyjoker992, line 203: I think we could create a private function for `placeBid()` with flag `isFromFeralFile` additionally so that we could reuse most of code between `placeBid()` and `placeSignedBid`. The difference is only validate the signature.

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,334 @@
...
+        );
+
+        require(
+            !auction_.isSettled,
+            "FeralfileEnglishAuction: auction already settled"
+        );
+
+        auctions[auctionID_].isSettled = true;
+
+        // saleData_.price = highestBid_.amount;
```
jollyjoker992: Why is it commented?

contracts/FeralfileEnglishAuction.sol
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    struct Bid {
+        address bidder;
+        uint256 amount;
+        bool fromFeralFile;
+    }
+
+    struct LatestAuctionStatus {
+        Bid latestBid;
```
jollyjoker992: The struct could be 
```
struct AuctionStatus {
        Bid highestBid
        uint256 endAt;
        bool isSettled;
}
```

contracts/FeralfileEnglishAuction.sol
tienngovan: accept finalize auction with price is zero

jollyjoker992 (approved):

PR #29, opened 2024-01-08
Contract for token ERC-1155 airdrop

jollyjoker992:
# The contract requirements. 
- Support both fungible and non-fungible. 
- Allow to pre-mint to contract itself using `mint(tokenID_, amount_)`, requires authorized call. 
- Allow to airdrop a tokenID to a list of addresses using `airdrop(tokenID_, to_)`, one address can receive only one token, requires authorized call. 
- Allow to burn remaining airdrop tokens.
- Allow owner to burn based on `burnable` flag. 
- Allow to stop the airdrop, no more token can be minted, no more airdrop.

### DIFF
--- a/contracts/FeralFileAirdropV1.sol
+++ b/contracts/FeralFileAirdropV1.sol
@@ -55,6 +55,12 @@ contract FeralFileAirdropV1 is ERC1155, Authorizable {
         _ended = true;
     }
 
+    /// @notice set contract URI
+    /// @param uri_ contract URI
+    function setURI(string memory uri_) external onlyOwner {
+        _setURI(uri_);
+    }
+
     /// @notice Mint tokens to the contract
     /// @param tokenID_ token ID
     /// @param amount_ amount of tokens to mint
--- a/contracts/FeralFileAirdropV1.sol
+++ b/contracts/FeralFileAirdropV1.sol
@@ -33,7 +33,7 @@ contract FeralFileAirdropV1 is ERC1155, Authorizable {
 
     // @notice end the airdrop
     function end() external onlyOwner {
-        require(_ended == false, "FeralFileTokenAirdrop: already ended");
+        require(!_ended, "FeralFileAirdropV1: already ended");
         _ended = true;
     }
 
@@ -41,11 +41,11 @@ contract FeralFileAirdropV1 is ERC1155, Authorizable {
     /// @param tokenID token ID
     /// @param amount_ amount of tokens to mint
     function mint(uint256 tokenID, uint256 amount_) external onlyAuthorized {
-        require(_ended == false, "FeralFileTokenAirdrop: already ended");
+        require(!_ended, "FeralFileAirdropV1: already ended");
         require(
             (tokenType == Type.Fungible && amount_ > 0) ||
                 (tokenType == Type.NonFungible && amount_ == 1),
-            "FeralFileTokenAirdrop: amount mismatch"
+            "FeralFileAirdropV1: amount mismatch"
         );
         _mint(address(this), tokenID, amount_, "");
     }
@@ -57,11 +57,11 @@ contract FeralFileAirdropV1 is ERC1155, Authorizable {
         uint256 tokenID,
         address[] calldata to_
     ) external onlyAuthorized {
-        require(_ended == false, "FeralFileTokenAirdrop: already ended");
+        require(!_ended, "FeralFileAirdropV1: already ended");
         require(
             (tokenType == Type.Fungible && to_.length > 0) ||
                 (tokenType == Type.NonFungible && to_.length == 1),
-            "FeralFileTokenAirdrop: amount mismatch"
+            "FeralFileAirdropV1: amount mismatch"
         );
 
         uint256[] memory _tokenIDs = new uint256[](1);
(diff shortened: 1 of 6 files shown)


contracts/FeralFileAirdropV1.sol
```diff
@@ -0,0 +1,133 @@
...
+        require(!_ended, "FeralFileAirdropV1: already ended");
+        _checkProperTokenAmount(amount_);
+        _mint(address(this), tokenID_, amount_, "");
+    }
+
+    /// @notice airdrop tokens to a list of addresses
+    /// @param tokenID_ token ID
+    /// @param to_ list of addresses to airdrop to
+    function airdrop(
+        uint256 tokenID_,
```
lpopo0856, line 77: suggestion: tokenId_

contracts/FeralFileAirdropV1.sol
```diff
@@ -0,0 +1,133 @@
...
+            uint256[] memory _amounts = new uint256[](1);
+            _amounts[0] = 1; // 1 token for each address
+
+            _safeBatchTransferFrom(
+                address(this),
+                to_[i],
+                _tokenIDs,
+                _amounts,
+                ""
+            );
```
lpopo0856: could we use `_safeTransferFrom` I think it will save some gas.

contracts/FeralFileAirdropV1.sol
```diff
@@ -0,0 +1,133 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import "@openzeppelin/contracts/token/ERC1155/ERC1155.sol";
+import "./Authorizable.sol";
+
+contract FeralFileAirdropV1 is ERC1155, Authorizable {
```
lpopo0856, line 7: `FeralfileAirdropV1` to fit the old naming

contracts/FeralFileAirdropV1.sol
```diff
@@ -0,0 +1,133 @@
...
+    enum Type {
+        Fungible,
+        NonFungible
+    }
+
+    // Version
+    string public constant version = "v1";
```
lpopo0856:
```solidity
    // version code of contract
    string public constant codeVersion = "FeralfileAirdropV1";
```

contracts/FeralFileAirdropV1.sol
```diff
@@ -0,0 +1,133 @@
...
+pragma solidity ^0.8.13;
+
+import "@openzeppelin/contracts/token/ERC1155/ERC1155.sol";
+import "./Authorizable.sol";
+
+contract FeralFileAirdropV1 is ERC1155, Authorizable {
+    enum Type {
+        Fungible,
+        NonFungible
+    }
```
lpopo0856, line 11: I'm not sure if keeping the strict Fungible/NonFungible rule is good. For example what if we have an activation that contains multi edition and multi-unique edition

jollyjoker992: I dont think it should happen. If it happened, we need to create v2, we cannot assume that.

contracts/FeralFileAirdropV1.sol
jollyjoker992: This is a new contract dedicated for the activation so i don't follow what we did for the exhibition contract. I think the `version` is more common to describe contract version and the `v1` is good enough as well.

contracts/FeralFileAirdropV1.sol
jollyjoker992: Why so? The `Feralfile` was a naming mistake, i think.

contracts/FeralFileAirdropV1.sol
lpopo0856: but we use that column to check the version at ff backend, missing that will make ff not able to identify this contract

contracts/FeralFileAirdropV1.sol
```diff
@@ -0,0 +1,126 @@
...
+    // Ended flag
+    bool private _ended;
+
+    constructor(
+        Type tokenType_,
+        string memory tokenURI_,
+        string memory contractURI_,
+        bool burnable_,
+        bool bridgeable_
+    ) ERC1155(tokenURI_) {
```
tienngovan, line 40: Do we need to provide an updating tokenURI function to change it in case something goes wrong or needs updating?

tienngovan: Do we need to provide function `uri` to return tokenURI?
I see this one on PETROLEUM contract
https://etherscan.io/address/0xedc227ad7fef4a2c6dbc2303a5732087314a4ce9#code

contracts/FeralFileAirdropV1.sol
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    // @notice end the airdrop
+    function end() external onlyOwner {
+        require(!_ended, "FeralFileAirdropV1: already ended");
+        _ended = true;
+    }
+
+    /// @notice set contract URI
+    /// @param uri_ contract URI
```
tienngovan: it's tokenURI, not contractURI

contracts/FeralFileAirdropV1.sol
jollyjoker992: That function is default by ERC1155.

PR #36, opened 2024-03-11
Feature/owner data contract

jollyjoker992:


contracts/OwnerData.sol
```diff
@@ -0,0 +1,319 @@
...
+        return result;
+    }
+
+    /// @notice remove data by indexes
+    /// @param contractAddress_ - the address of the contract
+    /// @param tokenID_ - the token ID
+    /// @param indexes_ - the indexes of the data to remove
+    function remove(
+        address contractAddress_,
+        uint256 tokenID_,
```
anhnguyenbitmark, line 171: Check if the `tokenID_` is the public token. We only reserve the right to delete records on public tokens.

tienngovan: fixed

anhnguyenbitmark (approved):

PR #39, opened 2024-04-11
Updated FFV4 supporting artist  and curator advance payment

tienngovan:


contracts/FeralfileArtworkV4_1.sol
```diff
@@ -0,0 +1,147 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import "./FeralfileArtworkV4.sol";
+
+contract FeralfileExhibitionV4_1 is FeralfileExhibitionV4 {
+    struct AdvancedData {
```
jollyjoker992, line 7: Should be `advance` (no need `ed` that could be wrong). Please check all related things as well.

contracts/FeralfileArtworkV4_1.sol
```diff
@@ -0,0 +1,147 @@
...
+        address artist;
+        uint256 amount;
+    }
+
+    mapping(address => uint256) public artistAdvancedAmounts;
```
jollyjoker992: - Should be advance (no need ed that could be wrong). Please check all related things as well.
- We shouldn't use `artist` specifically. We should treat it just like an address has been advanced, nothing more, so this variable generally serves for any addresses, no specific logic just for the artist. 
- We need to support update the address since the royalty address could be changed over time. Introduce a new function to replace the existing royalty address for this map is a need.

contracts/FeralfileArtworkV4_1.sol
```diff
@@ -0,0 +1,147 @@
...
+        string memory symbol_,
+        bool burnable_,
+        bool bridgeable_,
+        address signer_,
+        address vault_,
+        address costReceiver_,
+        string memory contractURI_,
+        uint256[] memory seriesIds_,
+        uint256[] memory seriesMaxSupplies_,
+        AdvancedData[] memory data_
```
jollyjoker992: I'd suggest passing data with the array of primitive type, i don't think create a struct just serve for the constructor is efficient in term of gas fee and also programming. We could use 2 different arrays like `advanceAddresses` and `advanceAmounts` and parse them by index respectively.

contracts/FeralfileArtworkV4_1.sol
```diff
@@ -0,0 +1,147 @@
...
+            signer_,
+            vault_,
+            costReceiver_,
+            contractURI_,
+            seriesIds_,
+            seriesMaxSupplies_
+        )
+    {
+        for (uint256 i = 0; i < data_.length; i++) {
+            artistAdvancedAmounts[data_[i].artist] = data_[i].amount;
```
jollyjoker992: Should validate the address is payable and the amount is greater than zero. Zero means no advance anymore.

contracts/FeralfileArtworkV4_1.sol
```diff
@@ -0,0 +1,147 @@
...
+            // send NFT
+            _safeTransfer(
+                address(this),
+                saleData_.destination,
+                saleData_.tokenIds[i],
+                ""
+            );
+            // distribute royalty
+            RevenueShare[] memory revenueShares = saleData_.revenueShares[i];
+            bool isArtist = false;
```
jollyjoker992: Those variable `isArtist` and the `artistAddr` could be unified as one, it looks like redundant. They serve for the same purpose, i think.

contracts/FeralfileArtworkV4_1.sol
```diff
@@ -0,0 +1,147 @@
...
+            address artistAddr;
+            for (uint256 j = 0; j < revenueShares.length; j++) {
+                if (artistAdvancedAmounts[revenueShares[j].recipient] > 0) {
+                    isArtist = true;
+                    artistAddr = revenueShares[j].recipient;
+                    break;
```
jollyjoker992: This is the assumption that we only pay advance for one address. Since the royalty addresses passed as the `calldata`, so there is a chance multiple royalty addresses need to be paid the advance. In the contract, we should support that even the backend doesn't need. We also need to define the priority when setup the advance map so that we can rely on the priority to pay the advance in case there is multiple addresses need to be paid once. 
For example, we need to pay the advance for address A, B and C. The priorities are 1, 1 and 2. That means if we need to pay the advance, we should consider A and B as the first accounts to be paid then it goes to the account C.  That would be complicated but it will be more logical.

contracts/FeralfileArtworkV4_1.sol
```diff
@@ -0,0 +1,160 @@
...
+        uint256[] calldata amounts_
+    ) public onlyAuthorized {
+        if (addresses_.length != amounts_.length) {
+            revert InvalidAdvanceAddressesAndAmounts();
+        }
+        for (uint256 i = 0; i < addresses_.length; i++) {
+            if (addresses_[i] == address(0)) {
+                revert InvalidAdvanceAddress();
+            }
+            if (amounts_[i] == 0 && advances[addresses_[i]] == 0) {
```
jollyjoker992: This check look redundant `advances[addresses_[i]] == 0`.

contracts/FeralfileArtworkV4_1.sol
```diff
@@ -0,0 +1,160 @@
...
+                revert InvalidAdvanceAmount();
+            }
+            if (amounts_[i] > 0) {
+                advances[addresses_[i]] = amounts_[i];
```
jollyjoker992: This will replace the advance item if it exists yet. We could check if the advance address exists, we shouldn't allow to update the amount, we just should allow to replace the address.

contracts/FeralfileArtworkV4_1.sol
```diff
@@ -0,0 +1,184 @@
...
+                ""
+            );
+            // distribute royalty
+            RevenueShare[] memory revenueShares = saleData_.revenueShares[i];
+            uint256 remainingRev = itemRevenue;
+
+            // deduct advances payment from revenue
+            for (uint256 j = 0; j < revenueShares.length; j++) {
+                if (
+                    advances[revenueShares[j].recipient] > 0 && remainingRev > 0
```
jollyjoker992: This could be simplified 
```
for (uint256 j = 0; j < revenueShares.length && remainingRev > 0; j++) {
       uint256 remainingAdvanceAmount = advances[revenueShares[j].recipient];
       uint256 rev = remainingAdvanceAmount >= remainingRev ? remainingRev : remainingAdvanceAmount;
       platformRevenue += rev;
       advances[revenueShares[j].recipient] -= rev;
       remainingRev -= rev;
}
```

test/feralfile_exhibition_v4_1.js
```diff
@@ -0,0 +1,1196 @@
...
+            web3.utils.toWei("0.5", "ether"),
+            web3.utils.toWei("1", "ether"),
+        ];
+
+        // 2. Update advance setting
+        try {
+            await contract.setAdvanceSetting(updatedAddresses, updatedAmounts);
+        } catch (error) {
+            console.log(error);
+            assert.equal(error.reason, "Custom error (could not decode)");
```
jollyjoker992, line 980: Can we detect the error specifically?

test/feralfile_exhibition_v4_1.js
```diff
@@ -0,0 +1,1196 @@
...
+        const advanceAddresses = [accounts[3], accounts[4]];
+        const advanceAmounts = [
+            web3.utils.toWei("0.3", "ether"),
+            web3.utils.toWei("0.8", "ether"),
+        ];
+
+        // 1. Set advance setting
+        await contract.setAdvanceSetting(advanceAddresses, advanceAmounts);
+
+        const advanceAddress3 = await contract.advances(accounts[3]);
```
jollyjoker992: Should be 
```
const advanceAmount0 = await contract.advances(advanceAddresses[0])
assert.equal(advanceAmount0, advanceAmounts[0]);
```

test/feralfile_exhibition_v4_1.js
```diff
@@ -0,0 +1,1196 @@
...
```
jollyjoker992: Same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/39/files#r1572039131).

jollyjoker992 (approved):

PR #41, opened 2024-05-21
add feralfile exhibition v4_2 & feralfile token contracts

hvthhien:
Story link: https://github.com/bitmark-inc/feral-file/issues/2273


Makefile
```diff
@@ -21,16 +21,18 @@ merge: sol-merger-check
 	mkdir /tmp/sol-merger && mkdir ./contracts/sol-merger && \
 	sol-merger --export-plugin SPDXLicenseRemovePlugin ./contracts/FeralfileArtworkV2.sol /tmp/sol-merger && \
     sol-merger --export-plugin SPDXLicenseRemovePlugin ./contracts/FeralfileArtworkV3.sol /tmp/sol-merger && \
-    sol-merger --export-plugin SPDXLicenseRemovePlugin ./contracts/FeralfileArtworkV4_1.sol /tmp/sol-merger && \
+    sol-merger --export-plugin SPDXLicenseRemovePlugin ./contracts/FeralfileArtworkV4_2.sol /tmp/sol-merger && \
```
jollyjoker992, line 25: Should keep the `V4_1` since it will be the current using contract. The `4_2` would just be one-off.

Makefile
```diff
@@ -39,14 +41,16 @@ build-contract: check
 	jq -r ".abi" build/contracts/FeralfileExhibitionV2.json > ./build/FeralfileExhibitionV2.abi && \
 	jq -r ".bytecode" build/contracts/FeralfileExhibitionV3.json > ./build/FeralfileExhibitionV3.bin && \
 	jq -r ".abi" build/contracts/FeralfileExhibitionV3.json > ./build/FeralfileExhibitionV3.abi && \
-	jq -r ".bytecode" build/contracts/FeralfileExhibitionV4_1.json > ./build/FeralfileExhibitionV4.bin && \
-	jq -r ".abi" build/contracts/FeralfileExhibitionV4_1.json > ./build/FeralfileExhibitionV4.abi && \
+	jq -r ".bytecode" build/contracts/FeralfileExhibitionV4_2.json > ./build/FeralfileExhibitionV4.bin && \
```
jollyjoker992: We could keep both since we need those ABI to handle properly from `autonomy-vault`

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,188 @@
...
+        uint256 seriesId;
+        uint256 tokenId;
+        address owner;
+        uint256 tokenIndex;
+    }
+
+    mapping(uint256 => TokenInfo) public _tokenInfos; // tokenID => tokenInfo
+    mapping(uint256 => uint256) public _tokenIndexes; // tokenID => tokenIndex
+
+    string public tokenPlaceholderURL;
```
jollyjoker992: Those values are quite confusing. Can we make them as what we define in metadata like `tokenPlaceholderAnimationURI` or `tokenPlaceholderImageURI`? That would be easier to me.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,188 @@
...
+import { FeralfileExhibitionV4_1 } from "./FeralfileArtworkV4_1.sol";
+import { Base64 } from "@openzeppelin/contracts/utils/Base64.sol";
+
+contract FeralfileExhibitionV4_2 is FeralfileExhibitionV4_1 {
+
+    error TokenIDNotFound();
+    error MintNotEnabled();
+
+    struct TokenInfo {
+        string thumbnail;
```
jollyjoker992: Same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/41/files#r1611369888), could it be `imageURI`?

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,188 @@
...
+            vault_,
+            costReceiver_,
+            contractURI_,
+            seriesIds_,
+            seriesMaxSupplies_
+        )
+    {}
+    
+    /// @notice Mint new collection of Artwork
+    /// @dev the function iterates over the array of MintDataWithIndex to call the internal function mintArtworksWithIndex
```
jollyjoker992: Might be a wrong comment.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,188 @@
...
+        if (!_exists(tokenId)) {
+            revert TokenIDNotFound();
+        }
+
+        _tokenInfos[tokenId] = TokenInfo(thumbnail, parameters);
+    }
+
+    /// @notice _buildArtworkData returns an object of artwork which would push to the actually artwork
+    /// @param parameters - the parameters for building artwork data
+    function _buildArtworkData(bytes memory parameters)
```
jollyjoker992: This function looks like redundant since it's really simple. Maybe just use `Base64.encode(parameters)` is OK

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,188 @@
...
+    function _buildPreviewURL(bytes memory parameters)
+        private
+        view
+        returns (string memory)
+    {
+        if (parameters.length == 0) {
+            return tokenPlaceholderURL;
+        }
+
+        return _buildIframe(_buildArtworkData(parameters), tokenBaseURI);
```
jollyjoker992: `tokenBaseURI` is quite confusing since it actually indicates the token URI, not an artwork file URI we're thinking. 
To reduce the confusion, I think we should implement the `setTokenBaseURI` and throw error there. Instead, we could create a new variable to store `artworkFileURI` and allow to set by authororized accounts.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,188 @@
...
+        private
+        pure
+        returns (string memory)
+    {
+        return Base64.encode(parameters);
+    }
+
+    /// @notice _buildPreviewURL returns the preview url
+    /// @param parameters - the token paramters
+    function _buildPreviewURL(bytes memory parameters)
```
jollyjoker992: The same naming suggestion [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/41/files#r1611369888).

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,188 @@
...
+        if (parameters.length == 0) {
+            return tokenPlaceholderURL;
+        }
+
+        return _buildIframe(_buildArtworkData(parameters), tokenBaseURI);
+    }
+
+    /// @notice _buildThumbnailURL returns the preview url
+    /// @param thumbnail - the token thumbnail
+    function _buildThumbnailURL(string memory thumbnail)
```
jollyjoker992: The same naming suggestion [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/41/files#r1611369888).

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,188 @@
...
+            return tokenPlaceholderThubmnail;
+        }
+
+        return thumbnail;
+    }
+
+    /// @notice _buildIframe returns a base64 encoded data for ff-frame
+    /// @param artworkData - the artwork data which would bring into the artwork
+    /// @param iframeURI - the artwork URL to be loaded into the iframe
+    function _buildIframe(string memory artworkData, string memory iframeURI)
```
jollyjoker992: Minor. Those params should be more clear. `tokenParams` and `artworkFileURI`.

jollyjoker992 (approved):

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,339 @@
...
+        }
+
+        TokenInfo memory tokenInfo = _tokenInfos[tokenId];
+        _tokenInfos[tokenId] = TokenInfo(imageURI, parameters, coinAmount);
+
+        // mint token for the owner
+        uint256 mintAmount = coinAmount - tokenInfo.coinAmount;
+        if (mintAmount > 0) {
+            address owner = ownerOf(tokenId);
+            _erc20Token.mint(owner, mintAmount * 10**18);
```
jollyjoker992: Should call `decimal()` from token contract instead of hard-coding the `18` decimals.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,339 @@
...
+            revert TokenIDNotFound();
+        }
+
+        TokenInfo memory tokenInfo = _tokenInfos[tokenId];
+        _tokenInfos[tokenId] = TokenInfo(imageURI, parameters, coinAmount);
+
+        // mint token for the owner
+        uint256 mintAmount = coinAmount - tokenInfo.coinAmount;
+        if (mintAmount > 0) {
+            address owner = ownerOf(tokenId);
```
jollyjoker992, line 104: I'm confirming with Sean if we should mint to the current NFT owner or the first owner.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,339 @@
...
+    /// @param s_ - part of signature for validating parameters integrity
+    /// @param v_ - part of signature for validating parameters integrity
+    /// @param saleData_ - the sale data
+    function buyArtworks(
+        bytes32 r_,
+        bytes32 s_,
+        uint8 v_,
+        SaleData calldata saleData_
+    ) external payable override {
+        require(_selling, "FeralfileExhibitionV4: sale is not started");
```
jollyjoker992, line 120: Why don't we just call `super.buyArtwork` instead? The coin transfer part could be done after that, it's OK to do them separately. 
I know using the same loop could reduce a bit gas fee but it would be just a little bit. Copy the code from the parent contract would make it harder to maintain the shared logic between those contracts. Please consider.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,339 @@
...
+            }
+
+            TokenInfo memory tokenInfo = _tokenInfos[saleData_.tokenIds[i]];
+            coinsTransferAmount += tokenInfo.coinAmount;
+
+            emit BuyArtwork(saleData_.destination, saleData_.tokenIds[i]);
+        }
+
+        // Call the transfer function of the ERC20 contract
+        if (coinsTransferAmount > 0 && _erc20Token.balanceOf(address(this)) > coinsTransferAmount * 10**18) {
```
jollyjoker992: if possible, please create a private function to get exact amount of coin instead of putting multiple magic values. 
Reference [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/41/files#r1616986156).

jollyjoker992 (changes requested):
Can you ask some more test for the coin balance after purchasing and update token info?

contracts/FeralfileArtworkV4_2.sol
hvthhien: we cannot call `super.buyArtwork` since the function was defined as external in the base contract. To do this we might need to change the access modifier in the base contract.

PR #42, opened 2024-06-06
support new buy artworks function for John Gerrard show

hvthhien:
Story link: https://github.com/bitmark-inc/feral-file/issues/2476


contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,184 @@
...
+                saleData_.price == msg.value,
+                "FeralfileExhibitionV4: invalid payment amount"
+            );
+
+        bytes32 message = keccak256(
+            abi.encode(block.chainid, address(this), saleData_)
+        );
+
+        //check nonce
+        _useCheckedNonce(saleData_.destination, saleData_.nonce);
```
jollyjoker992, line 110: It should be after the signature is validated. Otherwise, we could waste a little bit gas if the signature is invalid/expired.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,184 @@
...
+    FeralfileSaleDataV2,
+    Nonces
+{
+    error TokenIDNotFound();
+    error FunctionNotSupported();
+
+    // vault contract instance
+    IFeralfileVaultV2 public immutable VAULT_V2;
+
+    mapping(uint256 => uint256) private seriesNextSaleTokenIds; // seriesID -> tokenID
```
jollyjoker992: I'd suggest the name could be `seriesNextPurchasableTokenIds`

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,184 @@
...
+        string memory symbol_,
+        bool burnable_,
+        bool bridgeable_,
+        address signer_,
+        address vault_,
+        address costReceiver_,
+        string memory contractURI_,
+        uint256[] memory seriesIds_,
+        uint256[] memory seriesMaxSupplies_,
+        uint256[] memory seriesNextSaleTokenIds_
```
jollyjoker992: Same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/42/files#r1628973931)

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        uint256 itemRevenue;
+        if (saleData_.price > saleData_.cost) {
+            itemRevenue =
+                (saleData_.price - saleData_.cost) /
+                saleData_.quantity;
+        }
+
+        uint256 nextSaleTokenId = seriesNextSaleTokenIds[saleData_.seriesID];
```
jollyjoker992: Same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/42/files#r1628973931)

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,184 @@
...
+                revert TokenIDNotFound();
+            }
+
+            nextSaleTokenId++;
+            if (ownerOf(tokenIdForSale) != address(this)) {
+                continue;
+            }
+
+            // send NFT
+            _safeTransfer(
```
jollyjoker992, line 140: There would be a case that the remaining purchasable tokens is not enough compare to the requested quantity, the tx would be failed eventually but it will cost much gas since we transferred the token and emitted event already. So i'd suggest to accumulate the purchasable tokens into an array then loop over them to transfer.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,184 @@
...
+            for (uint256 j = 0; j < revenueShares.length; j++) {
+                address recipient = revenueShares[j].recipient;
+                uint256 rev = (remainingRev * revenueShares[j].bps) / 10000;
+                uint256 totalRev = saleData_.quantity * rev;
+                if (recipient == costReceiver) {
+                    platformRevenue += totalRev;
+                    continue;
+                }
+                distributedRevenue += totalRev;
+                payable(recipient).transfer(totalRev);
```
jollyjoker992: Should accumulate into one tx for the same recipient.

contracts/FeralfileArtworkV4_2.sol
hvthhien: this is a rare case that only happens when out of tokens. but storing tokens in an array and performing an additional loop would cost some additional gas for each transaction. Since it's a rare case, so I think it would be more efficient to keep the transfer inside this loop.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,184 @@
...
+        bytes32,
+        bytes32,
+        uint8,
+        SaleData calldata
+    ) external payable override {
+        revert FunctionNotSupported();
+    }
+
+    /// @notice Event emitted when Artwork has been sold with the additional nonce
+    event BuyArtworkV2(address indexed buyer, uint256 indexed tokenId, uint256 nonce);
```
jollyjoker992: Should be `BuyBulkArtworks`

contracts/FeralfileSaleDataV2.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import {IFeralfileSaleDataV2} from "./IFeralfileSaleDataV2.sol";
+
+contract FeralfileSaleDataV2 is IFeralfileSaleDataV2 {
+    function validateSaleDataV2(SaleDataV2 calldata saleData_) internal view {
```
jollyjoker992, line 7: Should be `validateSaleData`.

contracts/FeralfileVaultV2.sol
```diff
@@ -0,0 +1,57 @@
...
+    mapping(bytes32 => bool) private _paidSale;
+
+    constructor(address signer_) ECDSASigner(signer_) {}
+
+    /// @notice pay for buyArtwork to a FFV4 contract destination.
+    /// @param r_ - part of signature for validating parameters integrity
+    /// @param s_ - part of signature for validating parameters integrity
+    /// @param v_ - part of signature for validating parameters integrity
+    /// @param saleData_ - the sale data
+    function payForSaleV2(
```
jollyjoker992, line 19: Should be `payForSale`

contracts/IFeralfileSaleDataV2.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import {IFeralfileSaleData} from "./IFeralfileSaleData.sol";
+
+interface IFeralfileSaleDataV2 is IFeralfileSaleData {
+    struct SaleDataV2 {
```
jollyjoker992, line 7: Should be `SaleData`

hvthhien: Since the `SaleData` name has already been used, we need a different name for this struct.

contracts/FeralfileVaultV2.sol
hvthhien: the payload is different from the old Vault, so I changed the name to reduce the mistake when calling.

contracts/FeralfileArtworkV4_2.sol
hvthhien: we only use 1 revenue shares array for all the tokens, so we don't need to `accumulate` anymore

contracts/FeralfileArtworkV4_2.sol
hvthhien: currently, the event is for each single token

jollyjoker992: @lpopo0856 What would you prefer? Single event for this bulk sale or multiple events for individual?

contracts/IFeralfileSaleDataV2.sol
jollyjoker992: Ok. I thought you could use separate interface but it seems like we have to reuse the `RevenueShare`

contracts/FeralfileVaultV2.sol
jollyjoker992: Unless it makes us confused, otherwise i dont see any reason to name it with the version suffix. 
This function belongs to separate contract `FeralfileVaultV2` that actually doesn't make us confused when using it.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,211 @@
...
+{
+    error TokenIDNotFound();
+    error FunctionNotSupported();
+    error SaleNotStarted();
+    error InvalidPaymentAmount();
+    error TotalBpsOver();
+    error InvalidAddress();
+
+    // vault contract instance
+    IFeralfileVaultV2 public vaultV2;
```
jollyjoker992, line 24: Same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/42/files#r1634594351). 
There is no another vault v1 in this contract so it doesn't make us confused, so it should be named as `vault` and `setVault` is clear enough and remove the confusion.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,211 @@
...
+        SaleDataV2 calldata saleData_
+    ) external payable {
+        if (!_selling) {
+            revert SaleNotStarted();
+        }
+
+        super._checkContractOwnedToken();
+        validateSaleDataV2(saleData_);
+
+        if (saleData_.payByVaultContract) {
```
jollyjoker992: This should be done after the signature validation so we could save a bit gas fee if the signature is invalid.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,211 @@
...
+        bytes32 r_,
+        bytes32 s_,
+        uint8 v_,
+        SaleDataV2 calldata saleData_
+    ) external payable {
+        if (!_selling) {
+            revert SaleNotStarted();
+        }
+
+        super._checkContractOwnedToken();
```
jollyjoker992: This is not enough for checking the available token for sale. We should compare with the sale data quantity.

jollyjoker992 (approved):

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,221 @@
...
+            bridgeable_,
+            signer_,
+            vault_,
+            costReceiver_,
+            contractURI_,
+            seriesIds_,
+            seriesMaxSupplies_
+        )
+    {
+        vaultV2 = IFeralfileVaultV2(payable(vault_));
```
lpopo0856, line 58: Minor: check equal length of seriesIds_ and seriesNextPurchasableTokenIds_

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,221 @@
...
+                payable(recipient).transfer(totalRev);
+            }
+        }
+
+        if (
+            saleData_.price - saleData_.cost <
+            distributedRevenue + platformRevenue
+        ) {
+            revert TotalBpsOver();
+        }
```
lpopo0856, line 202: It's a miss from long time ago that we should add a check for `saleData_.price > saleData_.cost`

lpopo0856: or we could discuss should we allow to submit a buy tx that cost larger than price.

contracts/FeralfileArtworkV4_2.sol
```diff
@@ -0,0 +1,221 @@
...
+        // deduct advances payment from revenue
+        for (uint256 j = 0; j < revenueShares.length && remainingRev > 0; j++) {
+            uint256 remainingAdvanceAmount = advances[
+                revenueShares[j].recipient
+            ];
+            uint256 rev = remainingAdvanceAmount >= remainingRev
+                ? remainingRev
+                : remainingAdvanceAmount;
+            uint256 totalRev = saleData_.quantity * rev;
+            platformRevenue += totalRev;
```
lpopo0856: we have some logic error her, discuss in slack

lpopo0856 (approved):

PR #44, opened 2024-12-13
3078 - Series indexer contract

lpopo0856:
**Description**

- Story link: https://github.com/bitmark-inc/feral-file/issues/3078

**Document of the flow design and/or figma design**

- https://github.com/bitmark-inc/feral-file/issues/3078#issuecomment-2499603479


contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,448 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/access/Ownable.sol" as Ownable;
+
+contract SeriesIndexer is Ownable.Ownable {
+    // Counter
+    uint256 private nextSeriesID = 1;
```
jollyjoker992: Can you use the `length` of existing series map so we don't need to maintain this counter?

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,448 @@
...
+    uint256 private nextArtistID = 1;
```
jollyjoker992: Same has above [comment](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/44#discussion_r1895216957) but for the artist map.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,448 @@
...
+
+    struct Series {
+        string metadata;
```
jollyjoker992: I suppose this field is for series metadata source, so it should store the uri so the name should be `metadataURI`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,448 @@
...
+        string contractTokenData;
```
jollyjoker992: Same as above [comment](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/44#discussion_r1895217854), it should be `tokenDataURI`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,448 @@
...
+        uint256[] artistIDs;
+    }
+
+    // Artist Management
+    mapping(uint256 => address) private artistIDToAddress;
```
jollyjoker992: Why do we need this map instead of using the address directly as the artist ID?

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,448 @@
...
+    mapping(address => uint256) private addressToArtistID;
+    mapping(uint256 => bool) private ownerRightsRevokedForArtistID;
```
jollyjoker992: I think this should be on series level instead of artist level. We could also provide a feature for revoking all series from an artist.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,448 @@
...
+
+    modifier seriesExists(uint256 seriesID) {
+        require(bytes(seriesDetails[seriesID].metadata).length > 0, "Series does not exist");
+        _;
+    }
+    // ------------------------
+    // Internal Helper Functions
+    // ------------------------
+
+    function _checkOwnerOrArtist(uint256 seriesID) internal view {
```
jollyjoker992: Should pass the address instead of assume the `msg.sender`. The caller should determine which one should be passed, that will be more extensible and readable. If we just want to stick with the `msg.sender`, we might need a better name like `_checkMsgSenderIsOwnerOrArtist`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,448 @@
...
+        if (msg.sender == owner()) {
+            require(ownerCanModifySeries(seriesID), "Owner rights revoked by all artists in series");
+        } else {
+            _checkArtist(seriesID);
+        }
+    }
+
+    function _checkArtist(uint256 seriesID) internal view {
```
jollyjoker992: The same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/44#discussion_r1895230188).

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,448 @@
...
+    mapping(uint256 => uint256[]) private artistPendingCoArtistRequests;// artistID => SeriesID[]
+    mapping(uint256 => mapping(uint256 => uint256)) private artistPendingCoArtistRequestIndex;// artistID => SeriesID => Index
+
+    // Events
+    event ArtistAddressUpdated(uint256 indexed artistID, address indexed oldAddress, address indexed newAddress);
+    event SeriesIndexed(
+        uint256 indexed seriesID,
+        uint256[] artistIDs,
+        string metadata,
+        string seriesContractTokenCID
```
jollyjoker992: I would prefer using something like `URI` instead of `CID` that we need to stick to ipfs. Using `URI` could be better for extensibility later on.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,448 @@
...
+    ) external onlyOwner returns (uint256[] memory) {
+        uint256 length = artistsArray.length;
+        require(
+            length == metadatas.length &&
+            length == tokenIDsMapCIDs.length,
+            "Array length mismatch"
+        );
+        require(length > 0, "No series data");
+        require(length <= 50, "Batch size too large"); // Prevent DOS attacks
+
```
jollyjoker992: What does this mean for prevent DOS attacks?

contracts/SeriesIndexer.sol
lpopo0856: this's not possible since the mapping in solidity is a simple hash table without key management.

contracts/SeriesIndexer.sol
lpopo0856: that mean if someone try to put really large number of list, the contract will deny to service. It's not quite of an attack I think.

lpopo0856: the words is removed anyway

jollyjoker992: So that could be out of gas since it could reach the block gas limit so i don't consider it as "denial of service"

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,712 @@
...
+        /**
+     * @dev Deletes a series and cleans up related data
+     */
+    function _deleteSeries(uint256 seriesID) 
+        internal 
+        seriesExists(seriesID) 
+        onlyOwnerOrArtist(seriesID) 
+    {
+        Series storage series = seriesDetails[seriesID];
```
jollyjoker992: Should use `memory`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,712 @@
...
+        if (!ownerCanModifySeries(seriesID)) {
+            revert AllArtistsRevokedOwnerRightsError(seriesID);
+        }
+        _validateArtistsNotRevoked(artistAddrs);
+
+        // Remove current artists
+        Series storage series = seriesDetails[seriesID];
+        for (uint256 i = 0; i < series.artistIDs.length; i++) {
+            uint256 artistID = series.artistIDs[i];
+            isArtistIDInSeries[seriesID][artistID] = false;
```
jollyjoker992: It could be redundant since we did it in the `_removeSeriesFromArtist`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,712 @@
...
+            _removeSeriesFromArtist(artistID, seriesID);
```
jollyjoker992: Why don't we call the function `_removeArtistFromSeries` as well if we tend to remove the artist from series?

jollyjoker992: nvm, you did it when deleting the storage below.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,712 @@
...
+        uint256 seriesID,
+        address proposedArtistAddr
+    ) external seriesExists(seriesID) onlyOwnerOrArtist(seriesID) {
+        if (proposedArtistAddr == address(0)) {
+            revert ZeroAddressNotAllowedError();
+        }
+        
+        _ensureArtistHasID(proposedArtistAddr);
+        uint256 proposedArtistID = addressToArtistID[proposedArtistAddr];
+
```
jollyjoker992, line 256: It looks like you don't check the proposed address has been already artist for this series.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,712 @@
...
+        uint256 artistID = addressToArtistID[msg.sender];
+        if (artistID == 0) {
+            revert NotAnArtistError(msg.sender);
+        }
+        if (!seriesPendingCoArtist[seriesID][artistID]) {
+            revert NotAPendingProposalError(seriesID, msg.sender);
+        }
+
+        Series storage series = seriesDetails[seriesID];
+        series.artistIDs.push(artistID);
```
jollyjoker992: This logic could be presented already in `_addArtistsToSeries`, could it be reused?

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,706 @@
...
+
+    /**
+     * @dev Approves owner rights for the calling artist
+     */
+    function approveOwnerRightsForArtist() external {
+        uint256 artistID = addressToArtistID[msg.sender];
+        if (artistID == 0) {
+            revert NotAnArtistError(msg.sender);
+        }
+        if (!ownerRightsRevokedForArtistID[artistID]) {
```
jollyjoker992: Redundant parentheses

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+    // ============ State Variables ============
+
+    // ID Counters
+    uint256 private nextSeriesID = 1;
+    uint256 private nextArtistID = 1;
+
+    // Artist Management
+    mapping(uint256 => Artist) private artists;
+    mapping(address => uint256) private addressToArtistID;
+    mapping(uint256 => BitMaps.BitMap) private artistOwnerRightsRevokedBitMap;
```
jollyjoker992: I'm not so sure about your idea but it looks like a redundant declaration. We just need something like `BitMaps.BitMap private artistOwnerRightsRevokedBitMap`. If we define a mapping of `artistID` and `Bitmaps`, we cannot leverage the advantages of Bitmap to manage bitwise flag efficiently.

jollyjoker992: We can just checking the availability by simply call `artistOwnerRightsRevokedBitMap.get(artistID)`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+
+    // Series Management
+    mapping(uint256 => Series) private seriesDetails;
+
+    // Series - Artist Relationship
+    mapping(uint256 => mapping(uint256 => bool)) private seriesArtistExist;
```
jollyjoker992: Why don't we just use `mapping(uint256 => Bitmaps.Bitmap)`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+
+    // Co-Artist Management
+    mapping(uint256 => mapping(uint256 => bool)) private seriesPendingCoArtist;
```
jollyjoker992: The same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/44#discussion_r1904885819).

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        }
+        if (seriesCount > 50) {
+            revert BatchSizeTooLargeError(seriesCount);
+        }
+    }
+
+    /**
+     * @dev Checks if the caller is a series artist
+     */
+    function _requireIsSeriesArtist(uint256 seriesID) internal view {
```
jollyjoker992: This should be named as `_requireCallerIsSeriesArtist(uint256 seriesID)`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        emit ArtistAddressUpdated(artistID, oldAddress, newAddress);
+    }
+
+    // ============ Public/External View Functions ============
+
+    /**
+     * @notice Returns `true` if at least one artist in the series
+     *         has not revoked the owner's right to modify the series.
+     */
+    function ownerCanModifySeries(uint256 seriesID) public view returns (bool) {
```
jollyjoker992: This function could be named as `hasUnrevokedArtist(uint256 seriesID)`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        _;
+    }
+
+    // ============ Public/External Functions: Series Management ============
+
+    /**
+     * @dev Creates a new series
+     */
+    function addSeries(
+        address[] calldata artistAddrs,
```
jollyjoker992: Can you make clear variable name instead of abbreviation? This is the external function that would be seen publicly so I would suggest using clear name.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        string calldata tokenIDsMapURI
+    ) external returns (uint256) {
+        return _createSeries(artistAddrs, metadataURI, tokenIDsMapURI);
+    }
+
+    /**
+     * @dev Batch creation of series
+     */
+    function batchAddSeries(
+        address[][] calldata artistsArray,
```
jollyjoker992: This name could be confusing because of the word `array`, would just `artists` or `seriesArtists` be better?

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        string[] calldata metadataURIs,
+        string[] calldata tokenIDsMapURIs
+    ) external returns (uint256[] memory) {
+        _validateBatchParameters(artistsArray.length, metadataURIs, tokenIDsMapURIs);
+
+        uint256[] memory createdSeriesIDs = new uint256[](artistsArray.length);
```
jollyjoker992: Maybe just `seriesIDs` is good enough.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+     */
+    function _createSeries(
+        address[] memory artistAddrs,
+        string memory metadataURI,
+        string memory tokenIDsMapURI
+    ) internal returns (uint256) {
+        if (artistAddrs.length == 0) {
+            revert NoArtistsForSeriesError();
+        }
+        if (msg.sender == owner()) {
```
jollyjoker992: Could write a private function `_isCallerOwner` for removing duplicated checks.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        uint256 artistID = addressToArtistID[msg.sender];
+        if (artistID == 0 || !seriesArtistExist[seriesID][artistID]) {
+            revert CallerNotASeriesArtistError(seriesID, msg.sender);
+        }
+    }
+
+    /**
+     * @dev Ensures all provided artist addresses have not revoked owner rights
+     */
+    function _validateArtistsNotRevoked(address[] memory artistAddrs) internal view {
```
jollyjoker992: Could be `_ensureNoArtistsRevokedOwnerRights`, that will be more precise.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+            }
+        }
+        
+        seriesArtistExist[seriesID][artistID] = false;
+    }
+
+    /**
+     * @dev Validates that artists can only add series that only have themselves as artist
+     */
+    function _validateOnlySelfAsArtist(
```
jollyjoker992: Maybe just `_ensureCallerIsOnlyArtist` is good enough.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        if (artistAddrs.length == 0) {
+            revert NoArtistsForSeriesError();
+        }
+        if (msg.sender == owner()) {
+            _validateArtistsNotRevoked(artistAddrs);
+        } else {
+            _validateOnlySelfAsArtist(artistAddrs);
+        }
+        uint256 seriesID = nextSeriesID++;
+        _validateMetadataAndTokenURI(metadataURI, tokenIDsMapURI);
```
jollyjoker992, line 493: Should call this before the `seriesID` increment.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+    uint256 private nextSeriesID = 1;
+    uint256 private nextArtistID = 1;
+
+    // Artist Management
+    mapping(uint256 => Artist) private artists;
+    mapping(address => uint256) private addressToArtistID;
+    mapping(uint256 => BitMaps.BitMap) private artistOwnerRightsRevokedBitMap;
+
+    // Series Management
+    mapping(uint256 => Series) private seriesDetails;
```
jollyjoker992: I'm wondering why isn't it just `series` according to the `artists` declaration above?

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+
+    // ============ State Variables ============
+
+    // ID Counters
+    uint256 private nextSeriesID = 1;
+    uint256 private nextArtistID = 1;
+
+    // Artist Management
+    mapping(uint256 => Artist) private artists;
+    mapping(address => uint256) private addressToArtistID;
```
jollyjoker992: Could be `artistAddressToID`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+    }
+
+    /**
+     * @dev Update existing series' metadata and token data
+     */
+    function _updateSeries(
+        uint256 seriesID,
+        string memory metadataURI,
+        string memory tokenIDsMapURI
+    ) internal seriesExists(seriesID) onlyOwnerOrArtist(seriesID) {
```
jollyjoker992, line 513: I'd preferring using the modifiers in public/external function to leverage the visibility of them, we also could prevent potential gas waste when calling internal function multiple times.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+
+        emit SeriesUpdated(seriesID, series.artistIDs, metadataURI, tokenIDsMapURI);
+    }
+
+    /**
+     * @dev Deletes a series and cleans up related data
+     */
+    function _deleteSeries(uint256 seriesID) 
+        internal 
+        seriesExists(seriesID)
```
jollyjoker992, line 528: Same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/44#discussion_r1904908874)

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        }
+        if (bytes(tokenIDsMapURI).length == 0) {
+            revert EmptyTokenMapURIError();
+        }
+    }
+
+    /**
+     * @dev Validates batch parameters for series creation
+     */
+    function _validateBatchParameters(
```
jollyjoker992: This is quite ambiguous to me, maybe `_validateSeriesWriteBatch` could be better. It describes this is for series write operation in batch.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        string[] memory tokenIDsMapURIs
+    ) internal pure {
+        if (seriesCount != metadataURIs.length || seriesCount != tokenIDsMapURIs.length) {
+            revert ArrayLengthMismatchError(
+                seriesCount,
+                metadataURIs.length,
+                tokenIDsMapURIs.length
+            );
+        }
+        if (seriesCount == 0) {
```
jollyjoker992: Should check these count first since it's cheaper. 
And using in-memory variables to avoid access array `length` multiple times, it could be better in readability and gas (not so sure)

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+            _deleteSeries(seriesIDs[i]);
+        }
+    }
+
+    // ============ Public/External Functions: Series Artist Management ============
+
+    /**
+     * @dev Allows an artist to remove themselves from a series
+     */
+    function removeSelfFromSeries(uint256 seriesID) external seriesExists(seriesID) onlyArtist(seriesID) {
```
jollyjoker992: ChatGPT suggests a name `resignFromSeries` that's pretty good to me, It emphasizes the voluntary, you could consider.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        for (uint256 i = 0; i < artistIDs.length; i++) {
+            if (artistIDs[i] == artistID) {
+                artistIDs[i] = artistIDs[artistIDs.length - 1];
+                artistIDs.pop();
+                break;
+            }
+        }
+
+        // Remove series from artist's list
+        _removeSeriesFromArtist(artistID, seriesID);
```
jollyjoker992: Why don't we just create a function like `_removeArtistFromSeries` to grab above logic and make this function more structured? 
BTW, this name `_unlinkSeriesFromArtist` suggested by chatGPT is pretty good to emphasize the function purpose.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+            series.contractTokenDataURI
+        );
+    }
+
+    /**
+     * @dev Updates artist list for a series (owner only)
+     */
+    function ownerUpdateSeriesArtists(
+        uint256 seriesID,
+        address[] calldata artistAddrs
```
jollyjoker992: Should use clear name instead of abbreviation.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        for (uint256 i = 0; i < series.artistIDs.length; i++) {
+            uint256 artistID = series.artistIDs[i];
+            _removeSeriesFromArtist(artistID, seriesID);
+        }
+        delete series.artistIDs;
+
+        // Add new artists
+        _addArtistsToSeries(seriesID, artistAddrs);
+
+        emit SeriesUpdated(
```
jollyjoker992: This event could be unclear. We just update the series artists so we don't need to fire the event with metadata URI or token data URI, I'd suggest to create new event if needed.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        address[] calldata artistAddrs
+    ) external seriesExists(seriesID) onlyOwner {
+        if (!ownerCanModifySeries(seriesID)) {
+            revert OwnerRightsRevokedForThisSeries(seriesID);
+        }
+        _validateArtistsNotRevoked(artistAddrs);
+
+        // Remove current artists
+        Series storage series = seriesDetails[seriesID];
+        for (uint256 i = 0; i < series.artistIDs.length; i++) {
```
jollyjoker992, line 233: This block of code is for removing artists from series, I'd suggest write a private/internal function to do that. The `ownerUpdateSeriesArtists` is basically call 2 internal functions to remove and add artists again.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+                artistIDs[i] = artistIDs[artistIDs.length - 1];
+                artistIDs.pop();
+                break;
+            }
+        }
+
+        // Remove series from artist's list
+        _removeSeriesFromArtist(artistID, seriesID);
+
+        emit SeriesUpdated(
```
jollyjoker992: Same [here](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/44#discussion_r1905029940). If you just want to fire a general event for series update, you might just want to pass the `seriesID` and the client should use the seriesID to query the details from the contract. Adding more unrelated parameters in the event could cost more gas and also make it more confusing.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+    }
+
+    // ============ Public/External Functions: Co-Artist Management ============
+
+    /**
+     * @dev Proposes a new co-artist for a series
+     */
+    function proposeCoArtist(
+        uint256 seriesID,
+        address proposedArtistAddr
```
jollyjoker992: Shouldn't use the abbreviation.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+            uint256 artistID = nextArtistID++;
+            artists[artistID].artistAddress = artistAddr;
+            addressToArtistID[artistAddr] = artistID;
+        }
+    }
+
+    /**
+     * @dev Adds a pending co-artist request
+     */
+    function _addPendingCoArtistRequest(uint256 artistID, uint256 seriesID) internal {
```
jollyjoker992: We could name it better like `_addCoArtistProposal`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        // Add to series tracking
+        seriesDetails[seriesID].pendingCoArtists.push(artistID);
+        seriesPendingCoArtistIndex[seriesID][artistID] =
+            seriesDetails[seriesID].pendingCoArtists.length - 1;
+    }
+
+    /**
+     * @dev Removes a pending co-artist request
+     */
+    function _removePendingCoArtistRequest(uint256 artistID, uint256 seriesID) internal {
```
jollyjoker992: This should be named as `_removeCoArtistProposal`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+    error EmptyMetadataURIError();
+    error EmptyTokenMapURIError();
+
+    // ============ Structs ============
+
+    struct Series {
+        string metadataURI;              // IPFS hash or similar identifier for series metadata
+        string contractTokenDataURI;     // Token-related data for the series
+        uint256[] artistIDs;          // List of artists associated with this series
+        uint256[] pendingCoArtists;   // List of pending co-artists
```
jollyjoker992: Maintaining a pending artist list within a series is not a good choice to me. The reason is the pending artist is just a temporary data acts as an intermediate stage for artist proposal. In other words, it doesn't really relate to the series. The series should have artist, metadata and token data, that's what we need for a series. When query the series from contract, we don't care about the co-artist since it's not official one. 
Instead, I'd suggest maintain a pending co-artist as separate data, out of the series.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+    }
+
+    struct Artist {
+        address artistAddress;           // Artist's wallet address
+        uint256[] seriesIDs;            // List of series associated with this artist
+        uint256[] pendingCoArtistSeries; // List of pending co-artist series
```
jollyjoker992: Similar to this [comment](https://github.com/bitmark-inc/feralfile-exhibition-smart-contract/pull/44#discussion_r1905087949).

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+
+        emit CoArtistConfirmed(seriesID, artistID);
+    }
+
+    // ============ Public/External Functions: Artist Management ============
+
+    /**
+     * @dev Revokes owner rights for the calling artist
+     */
+    function revokeOwnerRightsForArtist() external {
```
jollyjoker992: I'd suggest using the name `revokeContractOwnerRights`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        if (_isArtistOwnerRightRevoked(artistID)) {
+            return;
+        }
+        _setArtistOwnerRightRevokedBool(artistID, true);
+    }
+
+    /**
+     * @dev Approves owner rights for the calling artist
+     */
+    function approveOwnerRightsForArtist() external {
```
jollyjoker992: I'd suggest using the name `approveContractOwnerRights`

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * @notice Checks if the given `artistAddr` has revoked owner rights
+     */
+    function ownerRightsRevoked(address artistAddr) external view returns (bool) {
```
jollyjoker992: Could be `hasArtistRevokedOwnerRights` for more clear.
Also shouldn't use abbreviation for parameter.

contracts/SeriesIndexer.sol
```diff
@@ -0,0 +1,792 @@
...
+        uint256 artistID = addressToArtistID[artistAddr];
+        return _isArtistOwnerRightRevoked(artistID);
+    }
+
+    /**
+     * @notice Returns all artist IDs for a given series
+     */
+    function getSeriesArtistIDs(uint256 seriesID) external view returns (uint256[] memory) {
```
jollyjoker992, line 403: Can we consider to expose a function to `getSeries` instead of expose multiple one that just easily achieved the same thing by just one function?

lpopo0856: I prefer returning the basic data type instead of struct so it could be used easier by client.

contracts/SeriesIndexer.sol
lpopo0856: I put this into internal function since the check unit is per series so we can reuse the modifier. If we want the modifier only at external function level we'll need to create both single and plural unit modifiers. It's ok to me, just confirm that's what we want to do.

jollyjoker992 (approved):

